PR #29, opened 2021-05-05
manage farcasterd children

Lederstrumpf:
builds on #27 to kill child processes of `farcasterd` at the end of tests before killing `farcasterd`.


TheCharlatan:
Works for me :)

I have two further feature requests. The tests currently fail on my machine if there is no `~/.farcaster_node`. It would be nice if we could pass `-d .some_dir` to the tests too so we don't need to rely on its creation. You can then pass them to the deamons when calling `launch(...` . I'd also enjoy the more fine grained logging from the rest of the code instead of using println.

PR #32, opened 2021-05-12
Walletd WIP

TheCharlatan:
Only WIP, opening to give some visibility.


src/walletd/runtime.rs
```diff
@@ -0,0 +1,123 @@
...
+        // with `error!` macro by the controller. If we propagate error here
+        // this will make whole daemon panic
+        Ok(())
+    }
+}
+
+impl Runtime {
+    fn send_farcasterd(&self, senders: &mut Senders, message: request::Msg) -> Result<(), Error> {
+        senders.send_to(
+            ServiceBus::Msg,
```
zkao, line 67: This should be ServiceBus::Ctl, Msg is for the peer Msgs

PR #39, opened 2021-06-14
Fix farcaster-core dependencies

h4sh3d:
Fix `farcaster-core` dependencies to commit https://github.com/farcaster-project/farcaster-core/commit/7ccfe1bcd1eb1961a2f678e640b0ebc8ce215ab4.

This is done because `farcaster-core` will have some big refactoring and will be unstable for some time.

**This should not break the current build (I tested) as it uses the same commit (see Cargo.lock diff) as before, but doing `cargo update` will not update the revision for core anymore, and thus become safe.**


TheCharlatan (approved):
tACK 9c2bfc94b5b47569e634407bbefcb46af473d162

PR #45, opened 2021-06-30
extend integration test to maker-taker

Lederstrumpf:
hardcoded serialized offer atm - should be read dynamically. since successful execution calls return unit type, testing is restricted to errors passed up - no validation of successful execution results.


zkao:
thank you!

PR #47, opened 2021-07-08
Message flow: Farcaster/Wallet/Swap improvements

zkao:

PR #48, opened 2021-07-09
Update core library

h4sh3d:
This PR starts integrating new core design with the latest refactoring. Work in progress, the node currently does not compile because of the new 'wallet' design in core and the fact that it is needed in `swapd` for actions on the `commit/reveal` phase.

Core points to https://github.com/h4sh3d/farcaster-core/tree/feature/encoding you can see the draft on core at https://github.com/farcaster-project/farcaster-core/pull/59

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -690,26 +690,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e88a8acf291dafb59c2d96e8f59828f3838bb1a70398823ade51a84de6a6deed"
 
 [[package]]
-name = "farcaster_chains"
-version = "0.1.0"
-source = "git+https://github.com/farcaster-project/farcaster-core?branch=cherry2#ff69dbf0ec6981f146043a9ded2a640b15129cf3"
-dependencies = [
- "bitcoin",
- "farcaster_core",
- "hex",
- "monero",
- "strict_encoding",
- "thiserror",
-]
-
-[[package]]
 name = "farcaster_core"
 version = "0.1.0"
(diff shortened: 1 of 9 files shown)


h4sh3d:
cc @zkao @TheCharlatan 

Please have a look at `swapd` and `walletd` and the use fo `Wallet` named `CoreWallet` when imported. I think some operations in swapd might be moved in walletd for the commit/reveal scheme or we create two wallets, one with keys and one without keys but it knows how to validate/compute hash (commit). What do you think?

zkao:
swapd must be able to validate stuff, as it has to construct and evolve its state, even if walletd is trusted. These systems are much simpler to analyse when each agent just ensures by itself that it is in a given state and just fire valid protocol steps.

how would that work?

h4sh3d:
Validation does not requires access to the wallet's seed in almost every cases, I added the ability to create the wallet in core (https://github.com/farcaster-project/farcaster-core/pull/61) without keys to access those methods who don't need them. see `CoreWallet::new_keyless();`.

The node on this PR now compiles with the latest changes in core. Tell me when(if) you want this rebased onto something.

zkao:
@h4sh3d this looks good to me, however #47 conflicts with this, as it touches all daemons, as i was working on the msg flow

h4sh3d:
@zkao we should merge #47 first and then I rebase this one for final review IMO

zkao:
cool, I will stop working on #47 till it gets merged, will move into networking

zkao:
@h4sh3d, please have a look at it, I moved many requests from Ctl to Msg bus, plus I went though most of the message flow (verifying it, and fixing mistakes), so im sure it wont be trivial to merge

h4sh3d:
@zkao ready for review after a rebase on `main`

PR #49, opened 2021-07-15
Enable the generic peer messaging

zkao:
Uses a local fork of rust-internet2 not yet public


zkao:
Ready to review, but do not yet merge, need to publish all custom version of dependencies, so far using local path on Cargo.toml

h4sh3d:
LGTM

PR #56, opened 2021-07-28
Farcasterd retrieve key from wallet before launching peerd

zkao:
Added  general mechanism for doing callbacks, such as: 
```
MakeOffer (orig req) -> return GetKeys (req) -> 
                                                         (rep) Keys -> continue MakeOffer (orig req)
```
Mechanism uses early return to exit original request, into the fallback path
requesting the keys, uses a RequestId to identify what pending_request to call
back, using recursion to continue the pending_requests following up receiving Keys back from wallet, such that the return is from the original request -- just proceding with execution

### DIFF
--- a/src/farcasterd/runtime.rs
+++ b/src/farcasterd/runtime.rs
@@ -13,7 +13,12 @@
 // along with this software.
 // If not, see <https://opensource.org/licenses/MIT>.
 
-use crate::{Senders, rpc::request::{Keys, LaunchSwap, PubOffer, RequestId, Token}, swapd::swap_id, walletd::NodeSecrets};
+use crate::{
+    rpc::request::{Keys, LaunchSwap, PubOffer, RequestId, Token},
+    swapd::swap_id,
+    walletd::NodeSecrets,
+    Senders,
+};
 use amplify::Wrapper;
 use request::{Commit, Params};
 use std::ffi::OsStr;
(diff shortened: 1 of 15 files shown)


src/farcasterd/runtime.rs
```diff
@@ -513,39 +527,53 @@ impl Runtime {
...
                     ));
-                } else {
-                    info!("Already listening on {}", &remote_addr);
-                    let msg = format!("Already listening on {}", &remote_addr);
-                    report_to.push((Some(source.clone()), Request::Progress(msg)));
-                }
+
                 let peer = internet2::RemoteNodeAddr {
-                    node_id: self.node_id()?,
+                    node_id: self.node_ids()[0], // FIXME shouldnt be 0 here
```
h4sh3d, line 576: Does that FIXME should be handle now?

h4sh3d (approved):
LGTM

zkao:
its good for now, problematic when node has multiple ids

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

src/farcasterd/runtime.rs
zkao: #922ce4d reports an error if there are multiple node_ids stating multiple ids are not supported

TheCharlatan (approved):

PR #59, opened 2021-08-09
Update farcaster core and amplify

h4sh3d:
Update `farcaster_core` to point on https://github.com/farcaster-project/farcaster-core on `main` and `amplify` dependencies to the latest versions.

General update of `Cargo.lock` and fix of new method name on `amplify::Getters` derive macro in `service.rs`.


zkao:
Thank you for figuring that out!

PR #61, opened 2021-08-16
Syncer integration0

zkao:


src/syncerd/bitcoin_syncer.rs
```diff
@@ -304,22 +329,11 @@ impl Synclet for BitcoinSyncer {
...
-                            Some(bh) => Some(bh.to_vec()),
-                            None => none!(),
-                        };
-
-                        state.change_transaction(tx.txid.to_vec(), blockhash, tx.confirmations)
-                    }
                 }
-                trace!("pending events: {:?}", state.events);
-
+                if !state.events.is_empty() || i == 0 || i % 25 == 0 {
```
TheCharlatan, line 333: Surely this is interim to get some more data through, right?

zkao: the hack is just to make it run at i = 0, the 25 was just to get a trace log once in a while, to understand syncer is running

PR #92, opened 2021-09-09
Syncer refactor

zkao:


TheCharlatan:
I'm having trouble getting this to build with the latest farcaster core main branch. Does it also fix the broken syncer_state unit tests?

zkao:
@TheCharlatan, with the Cargo.lock I added, it should compile

zkao:
unit test find -> filter should be fixed, yes

TheCharlatan:
The unit test is still broken, because the encoding is fucked up now with the address addendums. I see your frustration though, so lets merge this, to keep the train rolling. I will have the syncer types in the node by tonight, then we hopefully wont have any of this raw type encoding nonsense anymore.

zkao:
The address subscription is still supported, but half-broken, as previously discussed, i just generalized a bit for allowing using the same code for both subscription or polling, the subscription being just a control gate on the middle to prevent unnecessary polling on no address change.

PR #93, opened 2021-09-13
Add functional test for syncer

TheCharlatan:


zkao:
i recommend working on top of zkao:syncer_refactor

PR #97, opened 2021-09-14
Functional tests for syncer

TheCharlatan:
By default the functional tests are not run when executing `cargo test`. To run them, run `cargo test -- --ignored`.

Before running the functional tests, start the docker containers first by running `docker-compose up` in the `tests` directory. 

Docker creates and mounts a volume in the `data_dir` directory. The owner of this directory is the root user, so ownership has to be changed before running the functional tests with `sudo chown -R $USER data_dir`.

Alternatively, the regtest setup can be run directly on the host, with the tests expecting the following endpoints:

- Bitcoind rpc cookie at `./tests/data_dir/regtest/.cookie`
- Bitcoind at `http://localhost:18443`
- Electrs at `tcp://localhost:50001`
- Monerod at `http://localhost:18081` run with arguments `--regtest --offline --fixed-difficulty 1`
- Two instances of Monero-wallet-rpc at `http://localhost:18083` and `htp://localhost:18084` run with arguments `--disable-rpc-login --wallet-dir wallets`


Cargo.toml
```diff
@@ -85,8 +85,9 @@ zmq = "0.9"
 tokio = { version = "1.9.0", features = ["full"] }
 # Coin clients
 electrum-client = { git = "https://github.com/TheCharlatan/rust-electrum-client", branch = "verboseTransactionGet" }
-monero-rpc = { git = "https://github.com/TheCharlatan/monero-rpc-rs", branch = "changes" }
-# monero-rpc = { path = "../monero-rpc-rs/" }
+bitcoincore-rpc = "0.13.0"
+# monero-rpc = { git = "https://github.com/TheCharlatan/monero-rpc-rs", branch = "changes" }
+monero-rpc = { path = "../monero-rpc-rs/" }
```
zkao, line 90: local path

zkao (changes requested):
just fix the local path, and i test it on my side

Cargo.toml
TheCharlatan: Yeah, I was too early with publishing it. Let me also write some instructions.

TheCharlatan:
I added some documentation and ensured that the tests are only run when explicitly mentioned by annotating them with `#[ignore]`.

src/syncerd/monero_syncer.rs
```diff
@@ -172,7 +180,7 @@ impl MoneroRpc {
         address_addendum: XmrAddressAddendum,
     ) -> Result<AddressNotif, Error> {
         // TODO: Get network type from configuration
-        let network = monero::Network::Stagenet;
+        let network = monero::Network::Mainnet;
```
zkao, line 183: is this on purpose?

src/syncerd/bitcoin_syncer.rs
```diff
@@ -198,10 +199,15 @@ impl ElectrumRpc {
                 Ok(tx) => {
                     debug!("Updated tx: {}", &tx_id);
                     let blockhash = tx.blockhash.map(|x| x.to_vec());
-                    state.change_transaction(tx.txid.to_vec(), blockhash, tx.confirmations)
+                    let confs = match tx.confirmations {
+                        Some(confs) => confs,
+                        None => 0,
```
zkao, line 204: `let confs = tx.confirmations.unwrap_or(0)`

PR #110, opened 2021-09-21
Monero network enum

TheCharlatan:


zkao:
im not convinced about using lnpbp::chains, we should use our own. (1) its a global config, for all swaps, and in principle each swapd could ran swaps on different pairs.

h4sh3d:
There is a `Network` enum in core: https://docs.rs/farcaster_core/0.1.0/farcaster_core/blockchain/enum.Network.html

Maybe that's what you want?

PR #115, opened 2021-09-22
Re-Activate GitHub workflow

TheCharlatan:


.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: Tests
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      -main
   pull_request:
-    branches: [ master ]
```
h4sh3d, line 7: Why do we remove this? Doesn't that disable running workflows on PR on `main`?

Otherwise LGTM

TheCharlatan: No value means it runs with any pull request - otherwise this workflow would not be running right now.

h4sh3d (approved):

PR #119, opened 2021-09-26
Verify DLEQ proofs after reveal

Lederstrumpf:
Based on https://github.com/farcaster-project/farcaster-core/pull/143


zkao:
This commit is reviewed since yesterday, but I'm not yet merging because it breaks the swap because it call unimplemented stuff from core

Lederstrumpf:
> This commit is reviewed since yesterday, but I'm not yet merging because it breaks the swap because it call unimplemented stuff from core

#120 builds on this and removes the dependence on personal `farcaster-core` version, so this can be merged if that is too.

PR #124, opened 2021-10-04
Split out DLEQ proof from parameter commit reveal

Lederstrumpf:


Lederstrumpf:
Closing to instead build on #119

zkao:
It looks there was some doubled code artefact including an old bug of mine in one of the versions, probably from some bad merge/rebase, I will push the fixes

zkao:
I confirm that on af194e1 I can complete the swap

Lederstrumpf:
@zkao thanks for the detailed logs!

The issue was that I was using an electrum server without @TheCharlatan's extension, so Alice would never pick up the `Lock Tx` since the ` transaction_get_verbose` wasn't implemented.

Confirm that the `Alice:Maker, Bob:Taker` swap also works for me until the end now 👯

PR #168, opened 2021-11-09
Simplify main readme

h4sh3d:
Move remote client in docker documentation, the `Usage` section is now closer to the top for readers going through the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 # Farcaster: cross-chain atomic swaps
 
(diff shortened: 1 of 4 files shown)


TheCharlatan (approved):

README.md
```diff
@@ -66,27 +64,9 @@ swap-cli info
 
 Run `help` command for more details about available commands.
 
-### Remote client usage
```
TheCharlatan, line 69: Why remove this here?

h4sh3d: All the docker usage now point to the same doc explaining both, it makes the readme a bit shorter and IMO the first use case now for remote client usage is within Docker, so I thought it makes sens like this.

PR #172, opened 2021-11-09
Syncer: namespace zmq sockets in functional test

TheCharlatan:
The interprocess port might not be freed in time before running the next test case. Ensure this is not a problem by namespacing the ports for different tests.

This is not a problem in the normal syncer, since the port is re-used for all tasks.

### DIFF
--- a/tests/functional.rs
+++ b/tests/functional.rs
@@ -84,7 +84,7 @@ fn bitcoin_syncer_block_height_test(polling: bool) {
     let address = bitcoin_rpc.get_new_address(None, None).unwrap();
 
     // start a bitcoin syncer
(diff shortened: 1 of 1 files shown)


tests/functional.rs
```diff
@@ -35,6 +35,8 @@ use farcaster_node::syncerd::types::{
     WatchAddress, WatchHeight, WatchTransaction,
 };
 
+use amplify;
```
h4sh3d, line 38: Do we need amplify if we don't use `s!` macro?

h4sh3d (approved):

tests/functional.rs
TheCharlatan: No, removed it.

PR #173, opened 2021-11-09
Individual tests atomic

h4sh3d:
Based on #172.

- Split tests atomically, no big wrapper functions
- Remove no-capture, if test fail stdout is printed anyway, cleaner logs for passed tests
- Run all tests (bitcoin and monero ones) in sequence within 1 unique thread


TheCharlatan (approved):
Nice work!

PR #181, opened 2021-11-12
Replace get verbose transaction

h4sh3d:
Replace the get verbose transaction call to Electrum server by 3 different calls allowing to extract the tx confirmations and the block hash.

### DIFF
--- a/src/syncerd/bitcoin_syncer.rs
+++ b/src/syncerd/bitcoin_syncer.rs
@@ -262,7 +262,7 @@ impl ElectrumRpc {
                             );
                             let mut state_guard = state.lock().await;
                             state_guard
-                                .change_transaction(tx_id.to_vec(), None, None)
+                                .change_transaction(tx_id.to_vec(), None, Some(0))
                                 .await;
                             drop(state_guard);
                             continue;
@@ -289,7 +289,7 @@ impl ElectrumRpc {
                                     );
                                     let mut state_guard = state.lock().await;
                                     state_guard
-                                        .change_transaction(tx_id.to_vec(), None, None)
+                                        .change_transaction(tx_id.to_vec(), None, Some(0))
                                         .await;
                                     drop(state_guard);
                                     continue;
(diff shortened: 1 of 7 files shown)


src/syncerd/bitcoin_syncer.rs
```diff
@@ -242,20 +242,72 @@ impl ElectrumRpc {
...
+                    {
+                        Ok(history) => history,
+                        Err(err) => {
+                            debug!(
+                                "error getting script history, treating as not found: {}",
+                                err
+                            );
+                            let mut state_guard = state.lock().await;
+                            state_guard
+                                .change_transaction(tx_id.to_vec(), None, None)
```
TheCharlatan, line 265: `script_get_history` errors if there is no confirmation yet, so this should be `change_transaction(tx_id.to_vec(), None, Some(0))`.

src/syncerd/bitcoin_syncer.rs
```diff
@@ -242,20 +242,72 @@ impl ElectrumRpc {
...
+                            let block = match self.client.block_header(confirm_height) {
+                                Ok(block) => block,
+                                Err(err) => {
+                                    debug!(
+                                        "error getting block header, treating as not found: {}",
+                                        err
+                                    );
+                                    let mut state_guard = state.lock().await;
+                                    state_guard
+                                        .change_transaction(tx_id.to_vec(), None, None)
```
TheCharlatan, line 292: This also should be change_transaction(tx_id.to_vec(), None, Some(0)).

TheCharlatan (commented):
Thank you! Just some small changes to the logic needed. It's a pity we can't do fewer than 3 calls, but at least this works.

src/syncerd/bitcoin_syncer.rs
h4sh3d: Done

src/syncerd/bitcoin_syncer.rs
h4sh3d: Done

TheCharlatan (approved):

TheCharlatan (approved):

PR #185, opened 2021-11-16
Clippy

TheCharlatan:
All patches here were found by running `cargo clippy --all --all-targets --all-features -- -D warnings`. This patch should be tested extensively before merging to make sure no regressions were introduced.

However, there remain two unpatched linter findings that warrant further discussion here.

1. We use a bunch of enums that have a large size difference between their possible values. We could either Box these, or add clippy exceptions for them.
2. There seems to be a problem with `#[display(Debug)]` / `derive(Display)` where we take a reference that is immediately dereffed.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+enum-variant-size-threshold=4000
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,8 +10,8 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  static_analysis:
-    name: Static Analysis
+  rust_format:
+    name: Rust Format
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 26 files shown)


TheCharlatan:
I decided to bump the max enum variant size to 4k. Seems too complicated to either start using trait objects everywhere, or boxing the type.

TheCharlatan:
I still don't understand the issue with the formatter reference. After all `impl fmt::Display` takes a mutable reference as an argument `fn fmt(&self, f: &mut fmt::Formatter<'_>)`. Might this be a false positive?

h4sh3d:
> I still don't understand the issue with the formatter reference. After all `impl fmt::Display` takes a mutable reference as an argument `fn fmt(&self, f: &mut fmt::Formatter<'_>)`. Might this be a false positive?

I'm getting a few other errors but I don't see this one, what version of clippy/rust are you using?

TheCharlatan:
Clippy version: `clippy 0.1.58 (d914f17 2021-11-16)`
Rustc: `rustc 1.58.0-nightly (d914f17ca 2021-11-16)`
Cargo: `cargo 1.58.0-nightly (2e2a16e98 2021-11-08)`

h4sh3d:
Oh, 

> Clippy version: `clippy 0.1.58 (d914f17 2021-11-16)` Rustc: `rustc 1.58.0-nightly (d914f17ca 2021-11-16)` Cargo: `cargo 1.58.0-nightly (2e2a16e98 2021-11-08)`

Oh, you should maybe stay in `stable` for now, nightly for clippy is somewhat weird sometimes.

h4sh3d (approved):

h4sh3d (approved):

Issue #190, opened 2021-11-17
swapd pending requests: add more semantics to indexing the pending request, while awaiting for the indexed event to dispatch the pending request


zkao:

zkao:
from PR #191:

>The pending request has been ok so far, the issue is we index them using the ServiceId. I ensured it prints errors if it inserts a new pending request that removes another one that must not be removed, so if we deviate from the expected run, at least we log it and can fix it. So during the swap execution pending_requests are being inserted and removed at different states, and to reuse it has been ok, because its already empty when we try to use it again, while the protocol is unrolling. The issue is: its hard to make sure it behaves correctly on edge cases. And the way its currently done it makes it really hard to reason. I opened #190 for tracking this.

TheCharlatan:
I think this can be closed now, right?

PR #191, opened 2021-11-17
swapd: wait sweep_monero_thr blocks before sweeping the monero

zkao:
reuses pending_request: its getting dangerous

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -14,8 +14,8 @@
 // If not, see <https://opensource.org/licenses/MIT>.
 
 use crate::syncerd::{
-    opts::Coin, Abort, HeightChanged, SweepAddress, SweepAddressAddendum, SweepXmrAddress,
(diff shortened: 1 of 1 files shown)


TheCharlatan:
It seems to be stuck when running this waiting for the bitcoin lock transaction.

TheCharlatan (approved):
Seems to work nicely. That PendingRequest handling is ok in my eyes.

zkao:
The pending request has been ok so far, the issue is we index them using the ServiceId. I ensured it prints errors if it inserts a new pending request that removes another one that must not be removed, so if we deviate from the expected run, at least we log it and can fix it. So during the swap execution pending_requests are being inserted and removed at different states, and to reuse it has been ok, because its already empty when we try to use it again, while the protocol is unrolling. The issue is: its hard to make sure it behaves correctly on edge cases. And the way its currently done it makes it really hard to reason. I opened #190 for tracking this.

PR #193, opened 2021-11-18
Swapd state improvement, and rm unnecessary from_height

zkao:


TheCharlatan (approved):
tACK, but needs format.

zkao:
> tACK, but needs format.

comment formating that emacs fmt ignores  should be fixed on 9b1266f

PR #199, opened 2021-11-18
Further clippy findings

TheCharlatan:


src/swapd/runtime.rs
```diff
@@ -1434,8 +1429,8 @@ impl Runtime {
                         }
                     }
                     Event::TransactionConfirmations(TransactionConfirmations {
-                        id,
-                        block,
+                        id: _,
```
zkao, line 1432: the above examples, but these ones on ignoring stuff, i dont like so much

zkao: I find it cleaner like this:
``` rust
                  Event::TransactionConfirmations(TransactionConfirmations {
                        block,
                        ..
                   }
```

TheCharlatan: I added this, because I thought you are destructuring everything to make  it easier for yourself later on. Changing this to `..` instead then.

h4sh3d (approved):

zkao (approved):

Issue #197, opened 2021-11-18
Break Msg:Reveal in RevealMaker and RevealTaker


zkao:
Handling that message is overwhelming because it includes Alice or Bob in roles of both Maker or Taker Revealing. That is 4 different cases to be handled for this message, and they are special cases, it was a bad decision to try to be so generic

Offending code:
https://github.com/farcaster-project/farcaster-node/blob/main/src/swapd/runtime.rs#L794-L979

zkao:
Actually this can be resolved straight at the level of the pattern matching without any further structural changes

zkao:
Basically break it in two leaves matching the following patterns `Msg:Reveal(Reveal::AliceParameters(params))`
and `Msg:Reveal(Reveal::BobParameters(reveal))`

zkao:
Then it becomes less overwhelming

zkao:
The other option is to add a if conditional to the pattern, and use to distinguish between maker and taker, which is possibly is a better option, because the logic at that phase surrounds maker and taker, not alice or bob

zkao:
#201  doesnt split the msgs as proposed in the comments, but does the job anyway

PR #201, opened 2021-11-18
Improve state for reveal

zkao:


zkao:
This PR is ready to review, @Lederstrumpf 

it does not break the Reveal msg handling yet #197, but it creates state predicate functions that makes it possible to do so.

Not feeling brave enough to handle the  Reveal msg now :| Have to sort it out on the diagram first

Properly implementing this might also fix the Bob maker.

Lederstrumpf (approved):
lgtm. ignoring clippy's complaint re. the unused new functions since they're used in #207.

PR #205, opened 2021-11-19
Monero address encoding

TheCharlatan:


h4sh3d (approved):
At some point we can discuss if the addendum structs should live in node or core. IIRC there is some in core but they are not used. Saying this for future, leaving them in node is better for now IMO.

PR #206, opened 2021-11-19
Syncer TaskId

TheCharlatan:
This was more an exercise for myself than actually yielding much benefit. Especially the `SyncerBlockHeight` type is a bit unwieldy. Should I split their implementation into separate commits an PRs?

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -15,7 +15,7 @@
 
 use crate::syncerd::{
     opts::Coin, Abort, HeightChanged, SweepAddress, SweepAddressAddendum, SweepSuccess,
-    SweepXmrAddress, TaskTarget, WatchHeight, XmrAddressAddendum,
+    SweepXmrAddress, TaskId, TaskTarget, WatchHeight, XmrAddressAddendum,
 };
 use std::{
     any::Any,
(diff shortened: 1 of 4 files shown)


zkao:
i rebased #207 here, and then we have issues with offer, format changed, it seems

h4sh3d:
> i rebased #207 on #206 (on a private branch), and then we have issues with offer, format changed, it seems
> 
> alices: daemon:
> 
> ```
> [2021-11-20T15:27:38Z ERROR microservices::esb::controller] ESB request processing error: Farcaster core errors: Consensus error: Incorrect magic bytes
> ```

That’s probably an offer serialized with to_hex instead of to_string, at least the last 5 errors like this one were that for me.

zkao:
ack

h4sh3d:
`SyncerBlockHeight` is, as you said, not giving benefits. Maybe an implementation that generalized over different concept such as dags, or... I don't know what, but this is not needed now and maybe not even the right place to be generic, so I think you can drop the `SyncerBlockHeight` but keep the other one. (or discard both, I don't have a strong opinion)

TheCharlatan:
I removed all the block height typing, just the TaskId remains.

h4sh3d (approved):

TheCharlatan:
Do not merge before #230

TheCharlatan:
Rebased, now that #230 is merged.

h4sh3d (approved):
utACK, but should not modify current behaviors

zkao (approved):
no brainer

PR #226, opened 2021-11-22
Monero syncer minor

zkao:


zkao:
I am reviewing the monero syncer code base and ended up with these little changes

TheCharlatan:
@zkao can you remove the last commit? I found it improves the sweeping performance and makes it more reliable. Seems like state is more reliably cleared in the monero-wallet-rpc if you open a wallet twice.

src/syncerd/monero_syncer.rs
```diff
@@ -235,41 +235,32 @@ async fn sweep_address(
     let wallet = wallet_mutex.lock().await;
     trace!("taking sweep wallet lock");
 
-    match wallet
+    while let Err(err) = wallet
```
zkao, line 238: @TheCharlatan please review especifically this! all rest didnt change

zkao: if wallet needs to be open twice, maybe we have to sleeep instead?

zkao:
i had some commits from last night that i forgot to push :(

src/syncerd/monero_syncer.rs
TheCharlatan: I tried that, it did not really improve things.

TheCharlatan: This is still not quite the logic it had before.  For one, the wallet is no longer opened after being successfully generated. Further, the problem is that the watch-only flag in the monero-wallet-rpc seems to not reliably be cleared in the wallet once it opens a spend wallet from a view-only wallet. I am yet to analyze this properly, I was too busy here with other things here so far. I specifically added two tries, because after a lot of trial and error it was the sweet spot between holding up the execution of the other monero-wallet-rpc calls in the syncer and getting the swap to work more reliably.

zkao: the logic is different than before because before there were 2 independant paths, now only one path.

now if it tries to open a wallet, if it fails to open the wallet, it creates a new wallet and retry to open.

before it had two independent paths and they were different which is not what you usually want

zkao: your probably missed the line 238 at the top:

`     while let Err(err) = wallet`

TheCharlatan: You are right, this actually works pretty good!

TheCharlatan (approved):
tACK 638a8b0

PR #230, opened 2021-11-23
Env config names and defaults, merge Chain and Network args

h4sh3d:
- Shortify configuration names
- Merge lnpbp Chain and Farcaster Network
- others minor changes


zkao:
the configuration chain is not used, afaik

zkao:
and shall be removed

h4sh3d:
The configuration network is used in the syncers.

TheCharlatan (approved):

zkao:
> The configuration network is used in the syncers. And will be used while making an offer in my next PR.

it should not be possible to use global arguments for syncers, because farcasterd might launch 10 different syncers

zkao:
running for different coins on different sub networks

TheCharlatan:
I think if I understood @h4sh3d correctly, the configuration is relegated back to whatever is in the offer in his next PR.

zkao:
@TheCharlatan, can u shed some light on this, since u worked on the syncers config? 

but my understanding is that chain shall not be a global config, isnt it?

h4sh3d:
Uhm, not really. I thought about a setup where if launched with --testnet farcasterd cannot do mainnet swap in parallel. Do we really want to mix everything (in term of network)?

TheCharlatan:
Then I misunderstood the purpose @h4sh3d  , I agree with @zkao then that farcasterd should give this argument to the syncer manually.

h4sh3d:
> Then I misunderstood the purpose @h4sh3d , I agree with @zkao then that farcasterd should give this argument to the syncer manually.

It does, but through the shared flags and not with two parallel values.

So we want to allow running swaps on different network with the same farcasterd instance?

zkao:
a Syncer is:

https://github.com/farcaster-project/farcaster-node/blob/183398d95eb9255fb1af547860c326b0fba6340c/src/service.rs#L104

h4sh3d:
Yes, obviously. So farcasterd will be able to manage a list of Syncers by tuple `(Coin, Network)`, if I do a `make --testnet btc xmr` it will check and lauch both testnet syncer for btc and xmr, if 1 minute after I do `make --mainnet btc xmr` it will lauch the mainnet ones, correct?

zkao:
The obvious thing: syncer config should be local (independent of whether farcasterd mixes networks).  <- this is a bug

Farcasterd mixing or not networks is an opinionated detail that has no urgency <- this is an opinion

zkao:
merging Coin with the constructs used in core would also be appreciated

zkao:
the hard bit is how to make this farcasterd options disappear, any thought, @h4sh3d and @TheCharlatan ?

`--electrum-server localhost:60001 --monero-daemon http://localhost:38081 --monero-rpc-wallet http://localhost:18083`

h4sh3d:
> merging Coin with the constructs used in core would also be appreciated

Yes, I'll open an issue to track that and include it in core.

h4sh3d:
In my last commit (WIP) I use the `config` flag in args to parse a `toml` file containing the syncer configuration.

src/farcasterd/runtime.rs
```diff
@@ -176,10 +176,10 @@ impl Runtime {
             return Err(Error::Syncer(SyncerError::InvalidConfig));
         }
         if !self.syncers.contains_key(&(coin_arb.clone(), network)) {
-            launch("syncerd", &[coin_arb.to_string(), network.to_string()])?;
+            launch("syncerd", &[coin_arb.to_string()])?;
         }
         if !self.syncers.contains_key(&(coin_acc.clone(), network)) {
-            launch("syncerd", &[coin_acc.to_string(), network.to_string()])?;
```
zkao: if i were you, i would leave the stuff that is written to the toml to something runtime independent, such as a map btw say 
```
(Monero, Testnet) -> "--monero-daemon http://localhost:38081 --monero-rpc-wallet http://localhost:18083"
(Bitcoin, Testnet) -> "--electrum-server localhost:60001"
```
Then if the config doesnt have it, it can't run the swap.

src/bin/syncerd.rs
```diff
@@ -32,29 +32,20 @@ extern crate log;
 use clap::Clap;
 
 use farcaster_node::syncerd::{self, Opts};
-use farcaster_node::Config;
-
-fn main() {
-    println!("syncerd: blockchain tasks management microservice");
+use farcaster_node::{Error, ServiceConfig};
 
+fn main() -> Result<(), Error> {
```
TheCharlatan, line 37: What's up with the return type?

src/config.rs
```diff
@@ -12,35 +12,98 @@
...
-            msg_endpoint: opts.msg_socket.into(),
-            ctl_endpoint: opts.ctl_socket.into(),
+            syncers: Some(Syncers::default()),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(crate = "serde_crate")]
+pub struct Syncers {
```
TheCharlatan, line 60: Syncere is too generic a name for my taste, especially since we have other syncer vars in the code. How about just `SyncerConfig`

src/bin/syncerd.rs
h4sh3d: What is the problem about returning an error?

TheCharlatan: This function doesn't return, right?

h4sh3d: The binary returns eventually, so exit code is not zero and error is printed. You can have main return result in rust :)

h4sh3d: When I mean eventually: it returns early if it miss some params to start

TheCharlatan: > The binary returns eventually, so exit code is not zero and error is printed. You can have main return result in rust :)

But which statement within this main function actually returns? I only see a panic case here, and no `?`, `return` or omitted `;`. So I'm guessing you are just returning for style? But even then, it seems inconsistent with `swapd.rs` and `peerd.rs`.

h4sh3d: Oh sorry, I guess the code that was there got removed, run fn can panic but it is handled by the expect so yes I have to removed that. (On the phone, didn’t open the file before now to check)

h4sh3d: Done in 9e3a5fef5cc0610d9b79137e86706d6a1b7ff378

src/config.rs
h4sh3d: Done in 9e3a5fef5cc0610d9b79137e86706d6a1b7ff378

h4sh3d:
Fixing clippy...

TheCharlatan (approved):
tACK

PR #249, opened 2021-11-26
Syncer: TransactionConfirmations returns Bitcoin transaction

TheCharlatan:


h4sh3d (approved):
Since there is `Some` or `None` for seeing the tx I see why you can simplify the `Option<Vec<_>>` into just `Vec<_>`. Good catch

PR #256, opened 2021-11-29
farcasterd: rm syncer from syncer_services

zkao:


zkao:
@TheCharlatan if u r ok with the additional commit of mine, please merge

PR #259, opened 2021-11-29
improve swapd state management 

zkao:


src/swapd/runtime.rs
```diff
@@ -838,39 +907,27 @@ impl Runtime {
...
-
-                        match self.state.swap_role() {
-                            SwapRole::Alice => {}
-                            SwapRole::Bob => {
-                                let addr = self
-                                    .state
-                                    .b_address()
-                                    .cloned()
-                                    .expect("address available at CommitB");
-                                let msg = format!(
```
TheCharlatan, line 857: This message is no longer pushed to the progress report. Is your goal to put the funding address and amount under a different cli endpoint? Like it stands currently I cannot drive the swap with just the cli.

zkao: ops, i guess i forgot to push

zkao: @TheCharlatan can you report precisely the exact conditions you don't get the stuff pushed to the progress report?

zkao: 6d26a61 is the correct commit i tested

TheCharlatan: No, you are right. Progress shows up as before. Maybe I forgot to pull, sorry!

TheCharlatan (approved):
tACK ce36f882e983adca541a42a1aee3fd266de04eb3

Lederstrumpf (approved):
tACK - thanks for getting all that convoluted `remote_commit` matching out!

PR #268, opened 2021-11-30
Create config file if not present

### LINKED ISSUE #261: Create default `farcasterd.toml` config if missing
When farcasterd is launched, if no config file is found it should create one with the default values.

h4sh3d:
Fix #261

When launching farcasterd, if no config find is found a new one is created with the default values.


TheCharlatan (approved):
tACK

PR #281, opened 2021-12-01
Log levels

### LINKED ISSUE #170: Log level usage
Redefine log level such as:
- INFO (-vv): easy to follow the swap with messages such as this step happen, send money here, swap complete, funds sent to
- DEBUG (-vvv): services spawned, transaction details, mempool and block seen

### LINKED ISSUE #257: log should distinguish different swapds
```
2021-11-29T13:06:30Z INFO  farcaster_node::swapd::runtime] Terminating swap<0x7c29d85b9626425b9902e0f9bc5f49387f8b53e46d0df650ea511eae04a98077
```

### LINKED ISSUE #279: Printed final monero address in logs is Stagenet
At lines https://github.com/farcaster-project/farcaster-node/blob/main/src/walletd/runtime.rs#L1126-L1127

we print the Monero corresponding address after recovering all secret, but the address is hardcoded for Stagenet.

h4sh3d:
Fix #279, fix #257, fix #170 

Modify logs to resolve cited issues.

- [x] Alice Maker, Bob taker logs success path (INFO level)
- [x] Bob Maker, Alice taker logs success path (INFO level)
- [x] Swap id as prefix in swapd INFO level logs

Cancel with refund/punish path have not been tested here, but I think most of the work will be reused already and it can come in a later PR.

### DIFF
--- a/src/farcasterd/runtime.rs
+++ b/src/farcasterd/runtime.rs
@@ -152,9 +152,9 @@ impl Stats {
     fn success_rate(&self) -> f64 {
         let Stats { success, failure } = self;
         let total = success + failure;
-        let rate = *success as f64 / (total as f64) * 100f64;
+        let rate = *success as f64 / (total as f64);
         info!(
-            "{}: Success({}) / Total({}) = {:>4.3}%",
+            "{}: Success({}) / Total({}) = {:>4.3}",
             "Swap success rate".bright_blue_bold(),
             success.bright_white_bold(),
             total.bright_white_bold(),
--- a/src/farcasterd/runtime.rs
+++ b/src/farcasterd/runtime.rs
@@ -152,11 +152,12 @@ impl Stats {
     fn success_rate(&self) -> f64 {
         let Stats { success, failure } = self;
         let total = success + failure;
-        let rate = *success as f64 / (total as f64);
+        let rate = *success as f64 / (total as f64) * 100f64;
         info!(
-            "Swap success rate: Success({}) / Total({})  = {}",
-            success,
-            total,
+            "{}: Success({}) / Total({}) = {:>4.3}%",
+            "Swap success rate".bright_blue_bold(),
+            success.bright_white_bold(),
+            total.bright_white_bold(),
             rate.bright_yellow_bold()
         );
         rate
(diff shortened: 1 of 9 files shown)


src/swapd/runtime.rs
```diff
@@ -1421,7 +1425,7 @@ impl Runtime {
                         block,
                         tx,
                     }) if self.state.swap_role() == SwapRole::Alice => {
-                        info!(
+                        trace!(
```
TheCharlatan, line 1428: I'd like this on debug

h4sh3d: Done in 60d4f8e

TheCharlatan (approved):
tACK

src/farcasterd/runtime.rs
```diff
@@ -152,11 +152,12 @@ impl Stats {
     fn success_rate(&self) -> f64 {
         let Stats { success, failure } = self;
         let total = success + failure;
-        let rate = *success as f64 / (total as f64);
+        let rate = *success as f64 / (total as f64) * 100f64;
```
zkao, line 155: if u multiple by 100 this is not a rate

src/service.rs
```diff
@@ -111,15 +111,15 @@ pub enum ServiceId {
     #[from]
     Peer(NodeAddr),
 
-    #[display("swap<{0:#x}>")]
```
zkao, line 114: this was one of the few places that was printing the swapd correctly, are you sure about this?

zkao (changes requested):
Let me merge this one, so that it doesnt mess up with my unmerged work

src/farcasterd/runtime.rs
h4sh3d: Yes, I change this

src/service.rs
h4sh3d: The swap if can be printed with its full version in the logs one it start if you want, and we can access it with cli ls

src/farcasterd/runtime.rs
h4sh3d: Done in cbdf583dca7c126c8fee095ce5afe6ca3eb6cf3b

src/service.rs
h4sh3d: Done in 83619a737880eefab230b74327a8b0d7f5ae8fcd

PR #284, opened 2021-12-02
Add parallel swap test

TheCharlatan:
Adds a functional tests that runs four swaps in parallel (2 with Bob maker, 2 with Alice maker).

### DIFF
--- a/tests/functional-swap.rs
+++ b/tests/functional-swap.rs
@@ -5,8 +5,10 @@ use bitcoincore_rpc::{Auth, Client, RpcApi};
 use std::ffi::OsStr;
 use std::io;
 use std::process;
-use std::{thread, time};
+use std::sync::Arc;
+use std::time;
 use sysinfo::{ProcessExt, System, SystemExt};
+use tokio::sync::Mutex;
 
 use std::collections::HashMap;
 use std::str;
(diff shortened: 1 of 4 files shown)


tests/functional-swap.rs
```diff
@@ -49,46 +59,158 @@ async fn swap_test_bob_maker() {
...
+        xmr_dest_wallet_name,
+        execution_mutex,
     )
     .await;
 
     cleanup_processes(farcasterd_maker, farcasterd_taker);
 }
 
+#[tokio::test]
+#[ignore]
```
h4sh3d, line 97: Maybe we want to add `#[timeout(600000)]`, a 10 minute timeout (or more). Just in case some tests hang. (`use ntest::timeout;`)

tests/functional-swap.rs
```diff
@@ -166,9 +310,10 @@ async fn run_swap(
         .unwrap();
 
     // run until bob has the btc funding address
-    let address = retry_until_bitcoin_funding_address(cli_bob_progress_args.clone());
+    let address = retry_until_bitcoin_funding_address(cli_bob_progress_args.clone()).await;
 
     // fund the bitcoin address
+    let lock = execution_mutex.lock().await;
     let amount = bitcoin::Amount::ONE_SAT * 100000150;
```
h4sh3d, line 317: Can use `Amount::from_sat()`

h4sh3d (approved):
Don't know how to remove the clippy needless borrow error, it can with new rust 1.57.0. For now let's just relax a bit clippy by removing ` -- -D warnings` at the end. We can also add `msrv = "1.55.0"` at the top of `.clippy.toml` file for ignoring future rules that would not support our MSRV.

tests/functional-swap.rs
TheCharlatan: zkao said the same thing yesterday ^^ . I will fix all the amounts once your progress PR is in, then I can finally consume the progress messages for them and don't need to do any of this constant amount handling.

zkao (approved):

tests/functional-swap.rs
TheCharlatan: 18f9662

h4sh3d (approved):

Issue #287, opened 2021-12-03
Connect a specific version of node to farcaster.dev for tests


h4sh3d:
Website [farcaster.dev](https://farcaster.dev) can take a list of public offers, a special version of the node should be able to push offers there (the "maker" version) and the standard version should have a command to list them in cli.
- [ ] List available offers in cli
- [ ] Push new list of available offers in farcaster.dev from special node version

h4sh3d:
As @zkao mentioned, if we are able to list offers from the cli there is no need to have a special node version, we can just script to list and update every x time.

I though about what the cli could look like and came up with this idea:
```
swap-cli ls <TYPE>
```
where `TYPE` is by defaults `swaps` with `ls swaps` being the current `ls` behavior and `ls offers` being the list of offers like:
```
---
- "Offer:Cke4ftrP5A71LQM2fvVdFMNR4gumFhEqm11111uMFq7cEPSbSQpvEiF11111TB9zym113GTvtvqfD1111114A4TULEXhypzp8nJ4WAGAZUdDkYai3i9KTgUHkGLs5zX3N1tA11111111111111111111111111111111111111111AfZ113XRBuJApN3x"
```

zkao:
`ls` stands for `list swaps`

why reuse?

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

h4sh3d:
I though `ls` was for `list`

zkao:
maybe we should call it ListSwaps, and the abbreviation is ls, and the other ListOffers ?

Lederstrumpf:
created https://github.com/farcaster-project/offer-refiller to address this. It's pushing public offers to the web frontend whenever the list of swaps offered is updated.

PR #303, opened 2021-12-05
cli: add command to list PublicOffer

Lederstrumpf:
Lists public offer ids. First step towards addressing #287.
Requires merge of https://github.com/farcaster-project/farcaster-core/pull/195.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -904,8 +904,9 @@ checksum = "d9435d864e017c3c6afeac1654189b06cdb491cf2ff73dbf0d73b0f292f42ff8"
 
 [[package]]
 name = "farcaster_core"
-version = "0.4.2"
-source = "git+https://github.com/Lederstrumpf/farcaster-core?branch=public_offer_id_strict_encoding#e04587685d775485aded5406a2c5fd2602575181"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c57e7c7e9ed41b54cdb6ee48a1cbed541ae9dbbe040069ac81431c4f1b1cbdcb"
 dependencies = [
  "amplify",
  "base58-monero",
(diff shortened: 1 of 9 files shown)


h4sh3d:
Do you need something else in core for this? Otherwise you can release `0.4.3` with the CI

zkao:
the public offers are in info, you can check how its done there

zkao:
https://github.com/farcaster-project/farcaster-node/blob/13529bef2a15179f4e427b347019e10bbd27bad9/src/farcasterd/runtime.rs#L634

Lederstrumpf:
I've comment the `list-offer-ids` command for now since we're just pushing the full list of public offers for the PoC and not doing any diffing. Long-term, we should switch to pushing just the IDs since they're more compact and then have the receiver request the base58 encoded offers for anything they're missing.

So now the CLI has only `list-swaps` and `list-offers`, where `list-offers` dumps the public offers as in https://github.com/farcaster-project/farcaster-node/blob/13529bef2a15179f4e427b347019e10bbd27bad9/src/farcasterd/runtime.rs#L634

h4sh3d (approved):
tACK 42fe818

zkao:
@Lederstrumpf can we call swap-cli ls still?

if not can u add that?

h4sh3d:
@Lederstrumpf maybe there is some sort of "alias" in clap you can use to make `list-swaps` also respond on `ls`

zkao:
yes, its called short

zkao:
can u please fix that and rebase?

zkao:
@h4sh3d

h4sh3d:
Now we can use `ls`, `ls-swaps` and `list-swaps` or `ls-offers` and `list-offers`.

zkao:
> ls, ls-swaps and list-swaps or ls-offers and list-offers

there is mixed semantics here: either ls means list or list-swaps, but not both in different contexts

zkao:
`ls --swaps`
`ls --offers`

or 

`ls-swaps`
`ls-offers`

or

`ls`
`lo`

I prefer the latter, because the gain in characters are minimal of the others

PR #311, opened 2021-12-06
Attempt to curryfy log for syncer

h4sh3d:
An attempt to see how we can improve the logging stack.

As @TheCharlatan pointed out, in `log` there is the `target` argument in macro where we should be able to change the prefix (default to rust module), but I don't manage to change the target correctly yet.


zkao:
i guess you can close over the swapid and role at swapd launch (=curry those values in that fn), and set a lazy constant @h4sh3d

PR #312, opened 2021-12-06
Functional swap test: Sleep to evade race condition

TheCharlatan:


h4sh3d:
tACK bf5cfab9ade3edeb601fcaeb750e1ff373f06ffb

The added commit is to make sure we are compatible with the latest wallet rpc image, otherwise the port may not be set correctly.

h4sh3d (approved):

zkao (approved):

PR #318, opened 2021-12-07
Get transaction

zkao:
@TheCharlatan


zkao:
blocking logic to be avoided

PR #335, opened 2021-12-09
docs: add sequence diagram svg

zkao:


h4sh3d (approved):
I think we can even include it at the end of doc/README.md, just after the list of docs.

Issue #320, opened 2021-12-07
expose finality threshold externally


zkao:
config or pass the parameters when making and taking? @h4sh3d @TheCharlatan @Lederstrumpf

Lederstrumpf:
I lean towards passing as parameters: for low exchange amounts, one may prefer lower finality thresholds.

Lederstrumpf:
Or one uses different configs in this case. Actually, I prefer that ^^

PR #175, opened 2021-11-10
Update finality threshold for testnet swaps

h4sh3d:
Testnet Bitcoin timelocks should be: 4 (cancel) and 6 (punish)

PR #339, opened 2021-12-10
Update finality threshold for btc/xmr

h4sh3d:
Replace #175 

Set btc to 1 block to consider final and xmr to 5 blocks (even if not spendable before 10).

This is temporary and will be changed towards #320.


TheCharlatan (approved):
It's amazing that we can just change these fundamental params and the swaps still work!

PR #340, opened 2021-12-10
Extract auto funding in config

h4sh3d:
Move the `--auto-fund` parameter from cmd line argument to `farcasterd.toml` config file and added list of node and paramter to use per network when auto-funding is enable.

The funding logic stays exactly the same, just where parameters come from has changed. The pattern matching has changed to include the network selection to see if auto-funding is configure or not.

**Auto-funding will work only if `auto_fund` is set to true and if the network is registered in the config, otherwise manual funding is used.**

### DIFF
--- a/farcasterd.toml
+++ b/farcasterd.toml
@@ -1,3 +1,24 @@
+# Farcasterd configuration
+# configures farcasterd specific behaviors such as auto-funding swaps
+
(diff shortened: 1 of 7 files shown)


TheCharlatan (approved):
Nice job @h4sh3d

PR #341, opened 2021-12-10
swapd: stop info logging after finality thr crossed + funding message INFOed only when safe to fund

zkao:


zkao:
please review the two commits independently

TheCharlatan (approved):
tACK

h4sh3d:
This has been included in #343

PR #349, opened 2021-12-10
Swapd: Lower xmr finality threshold to 1

TheCharlatan:
Bitcoin testnet is too fast at the moment, leading to all swaps being cancelled.
Should we put it to 0 even?


h4sh3d:
I think when btc goes too fast the cancel timelock should be increased instead of decreasing finality. With cancel set to 10 or 30 that should work no?!

TheCharlatan:
I just set cancel confs to 20 and it did not reach the required finality.

PR #352, opened 2021-12-11
farcasterd: list puboffers excluding the consumed offers

zkao:


zkao:
should resolve delay reported by @Lederstrumpf

PR #362, opened 2021-12-14
Swap test funding

TheCharlatan:


Lederstrumpf (commented):
@TheCharlatan: the functional tests are consistently timing out these days:
https://github.com/farcaster-project/farcaster-node/runs/4542481454?check_suite_focus=true

But also what's on `main` right now.

I think it may be related to this?
https://github.com/farcaster-project/farcaster-node/runs/4542481454?check_suite_focus=true#step:7:22819
`[2021-12-16T03:00:26Z DEBUG farcaster_node::syncerd::bitcoin_syncer] error getting transaction, treating as not found: Electrum server error: {"code":1,"message":"not indexed tx 14ff84272a83f1b875eead5b95a574c7e973efaaf2ffdf3dcf58e650e13789f4"}`

TheCharlatan:
No, definitely not related to that, only the parallel swaps are failing. I think the block generation is still not isolated enough between single swaps, so the swaps may enter states that are not expected yet. I moved the mutex now to act on the entire swap. Lets see if that improves it. I also clarified the `parallel` swap name a bit, since the swaps don't execute in parallel yet, it just makes and takes in parallel. I'm going to work on truly parallel swap tests soon.

Lederstrumpf (approved):

PR #363, opened 2021-12-14
Cancel funding

zkao:


src/swapd/runtime.rs
```diff
@@ -1691,7 +1691,17 @@ impl Runtime {
                             );
                             return Ok(());
                         }
-                        if let Some(tx_label) = self.syncer_state.tasks.watched_addrs.remove(id) {
+                        if let Some(TxLabel::AccLock) =
+                            self.syncer_state.tasks.watched_addrs.remove(id)
```
TheCharlatan, line 1691: Why are you reverting this? This was what caused you problems and made bob think he had to complete monero funding.

PR #386, opened 2021-12-16
Reduce log level of Docker Compose and add container name declarations for each container

sethforprivacy:


sethforprivacy:
Also pushed a commit to add container_name declarations for each container, making them easier to manipulate with `docker` if not in the same directory as `docker-compose.yml`/

h4sh3d (approved):

PR #395, opened 2021-12-20
Fee estimation

TheCharlatan:
Based on #388

@zkao , I am loosing quite some precision with the amount parsing. The reason I'm using the bitcoin amount type is that sending a float in the emitted Event is kind of hard to get right, since we can't hash floats. Do you want a string instead, that can then be parsed into a float?

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -2413,6 +2413,9 @@ impl Runtime {
                     Event::TransactionRetrieved(event) => {
                         debug!("{}", event)
                     }
+                    Event::FeeEstimation(event) => {
+                        debug!("{}", event)
+                    }
                 }
             }
             Request::Protocol(Msg::CoreArbitratingSetup(core_arb_setup)) if self.state.reveal() => {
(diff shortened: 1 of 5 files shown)


zkao:
@TheCharlatan we need at least 2 decimals (tx sizes on the order of >100vB, thus 0.01 sat/vB * 100 vB = 1 sat)

zkao:
there was a bug in core that forced using sats for fees, though

but also theres some microsats unit or some other small unit, in rust-bitcoin

TheCharlatan:
Funny that you mention, because I just saw that in rust-bitcoin. Would require us to upgrade farcaster-core first though :/

TheCharlatan:
Actually, I think the precission is higher than you state here, since it is in kB.

h4sh3d:
Core uses rust Bitcoin amount which can’t store less than a sat internally.

TheCharlatan:
I think core should eventually use sat/kVByte as well.

zkao:
that would work too, as we have more than enough precision

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

zkao:
although i find that unit a bit counter intuitive, because txs are smaller than 1kB

zkao (approved):

zkao (approved):
what is the idea here:

- remove fee from offer and let syncer / swap / wallet pick the fee?
- just validate the fees?

zkao:
@TheCharlatan

PR #404, opened 2022-02-21
Align dependencies with lnp-node e81e693d1a92fc5aef10c423648db43e4b755e97

sedited:
The deps are alligned, but parallel swap test does not run yet. 

As part of this PR and besides the pure API changes arising from the deps bump I also changed:

- Peerd raw unencrypted connection to brontide encrypted connection
- `esb::SenderList`to `Endpoints`
- Renamed `senders` to the more appropriate `endpoints`
- Use `esb::Error<ServiceId>` instead of `esb::Error`
- Use lnp `Ping`, `Error` and `Init` messages instead of our own. This permanently binds us to the lnp::p2p crate


TheCharlatan:
Needs merge of https://github.com/farcaster-project/farcaster-core/pull/215

Lederstrumpf (approved):
tACK

TheCharlatan:
Requires upstream patching in either internet2 or microservices. Tracking in https://github.com/LNP-WG/lnp-node/issues/50

TheCharlatan:
Seems like the CI decided to pass now, so I'll open it for review.

h4sh3d:
Let me push 1 change so the monero functional tests pass with the newest `monero-lws:latest`. They currently don't because the container crash at startup because the name of a var changed.
After review and merge I can better test https://github.com/farcaster-project/farcaster-node/pull/444 and see if there is no real regression.

Cargo.toml
```diff
@@ -39,7 +39,9 @@ required-features = ["server"]
 
 [dependencies]
 # Farcaster crates
-farcaster_core = { version = "0.4.3", features = ["serde"] }
+# farcaster_core = { version = "0.4.4", features = ["serde"] }
+farcaster_core = { git = "https://github.com/farcaster-project/farcaster-core", branch = "main", features = ["serde"] }
```
Lederstrumpf, line 44: I just did a release of core so that we don't have this wart anymore, but it failed since we currently have our own fork of secp256kfun in there. But I'm good to merge without this fixed.

Lederstrumpf (approved):
tACK @ `0f38acb` - thanks for having all your teeth pulled @TheCharlatan! ;)

h4sh3d (approved):

TheCharlatan:
I can has merge?

PR #434, opened 2022-05-14
Syncer functional test: Test sweep minimum amount

### LINKED ISSUE #423: Syncer: Add unit test for minimum sweep balance enforcement
Functional syncer test currently only checks that sweep is triggered once the balance is non-zero. It should also check that it is not triggered before a minimum balance threshold on a certain address is reached. This is a follow-up for #417 .

TheCharlatan:
To test that the minimum amount is respected, first send an XMR amount below to required minimum to the sweep address, then send another amount to complete the required minimum. The syncer should only send a single event for the sweep of the entire amount, and not two events. A second, non-blocking receive on the zmq socket should therefore error. Closes #423 .


Lederstrumpf (approved):
tACK

PR #436, opened 2022-05-14
Cli: Fix make command failure reporting

TheCharlatan:
The failure was never reported before, because of an early return


TheCharlatan:
Rebased.

src/farcasterd/runtime.rs
```diff
@@ -857,59 +857,56 @@ impl Runtime {
...
-                        Request::Success(OptionDetails(Some(msg))),
-                    ));
-                    self.arb_addrs.insert(pub_offer_id, arbitrating_addr);
-                    self.acc_addrs
-                        .insert(pub_offer_id, monero::Address::from_str(&accordant_addr)?);
-                } else {
-                    let msg = "This Public offer was previously registered";
-                    warn!("{}", msg.err());
+                };
+                if res.is_err() {
```
h4sh3d, line 904: use `if let Err(err) = res {` instead and avoid `.err().unwrap()` later.

h4sh3d: use `if let Err(err) = res {` instead and avoid `.err().unwrap()` later.

h4sh3d (changes requested):
ACK

src/farcasterd/runtime.rs
TheCharlatan: Done in 241291397afca2d3b87c1c5abfbd86b4273bb000

h4sh3d (approved):

Issue #285, opened 2021-12-02
Make all cli messages serializable


TheCharlatan:
Output from the `peers`, `ls` and `info` commands are yaml serializable. Optimally, the `progress` messages should be too to make it easy for external applications to consume messages.

h4sh3d:
Uhm, the way progress currently works it is not possible, but what kind of yaml output would you expect? In other words what would be the structure?

TheCharlatan:
There's no point working on this now, it's going to take too much time. We will have to refactor the progress workflow, there's no really good reason why this shouldn't work. Lets get your proposal in and then we can think about how it can be yaml'ified.

TheCharlatan:
The `needs-funding` messages can be made serializable once the amount serialization is available in monero-rs.

TheCharlatan:
Only funding info remains. Funding info can be tackled once a new monero-rs version is released and bumped here with the new serde serialization on the amount type.

PR #437, opened 2022-05-15
Make offer / take offer cli yaml serializable

TheCharlatan:
Based on #436 , addresses parts of #285 .


h4sh3d:
In an ideal world we would have something like `--format yaml` in cli that allow printing yaml, with a default to "string" or human readable format.

At least for some cli actions like this one were both can be justified. What do you think?

TheCharlatan:
Sure, what really counts for me is that we use proper data structures and are consistent between the cli calls. We could also add `--format json` easily, or `--format string`. I think we should add this (in a separate series of PRs), since all we need to do is add some serde traits and slightly adjust the cli binary. Ideally such a code change would not touch the other, permanent, microservices at all (and I don't think it would).

src/farcasterd/runtime.rs
```diff
@@ -928,90 +927,98 @@ impl Runtime {
...
+                let daemon_service = internet2::RemoteNodeAddr {
+                    node_id,                                           // checked above
+                    remote_addr: RemoteSocketAddr::Ftcp(peer_address), /* expected RemoteSocketAddr */
+                };
+                let peer = daemon_service
+                    .to_node_addr(LIGHTNING_P2P_DEFAULT_PORT)
+                    .ok_or(internet2::presentation::Error::InvalidEndpoint)?;
+
+                // Connect
+                info!("going twice?????????????");
```
h4sh3d, line 954: Still needed?

TheCharlatan: No, must have snuck back in during rebase. Removed in 0e194d8d37b3cb973ff60176fa58285e02fd226a.

h4sh3d (approved):

PR #439, opened 2022-05-15
add userpass auth option for bitcoin rpc

Lederstrumpf:
Adds optional userpass authentication for bitcoin-rpc when autofunding. When both cookie and userpass authentication credentials are provided, it defaults to cookie authentication.


TheCharlatan (approved):
light code review ACK.

PR #442, opened 2022-05-23
State clean up

zkao:


TheCharlatan:
None of the swap functional tests pass, neither in the CI nor on my local machine. On main they run fine. The monero syncer address lws test is expected to fail until #448 is fixed by #444 .

Lederstrumpf (changes requested):
There's a failure in the functional tests induced by your change:
```
thread '<unnamed>' panicked at 'timeout before any monero funding address could be retrieved', tests/functional-swap.rs:1733:5
```

my _guess_ is that the xmr funding address is not part of swapd's state anymore (with your changes) when it's queried in the test. if this is correct, please figure out whether this should be the case (in which case change when the test queries), or if not, retain the funding address in the state at that stage of the swap.

zkao:
it works until [9ec26af](https://github.com/farcaster-project/farcaster-node/pull/442/commits/9ec26af096912e36d73079fe58f26f1b9b03cc4e) so the issue is local_params that is not present long enough somewhere and contains the destination address

sedited (approved):
tACK

PR #444, opened 2022-05-24
Bump CI container versions

h4sh3d:
Update the CI functional tests container versions:
- bump bitcoin to `22.0`
- bump monerod to `0.17.3.2`
- bump monero-wallet-rpc to `0.17.3.2`

Syncer and tests must be rewritten to work with bitcoin `23.0` and electrs `0.9.x`. This is out of scope of this PR.

### DIFF
--- a/tests/functional.rs
+++ b/tests/functional.rs
@@ -931,7 +931,7 @@ fn create_bitcoin_syncer(
         "--coin",
         "Bitcoin",
         "--electrum-server",
-        &format!("tcp://{}:50001", ehost),
+        &format!("tcp://{}:60401", ehost),
     ]));
 
     syncer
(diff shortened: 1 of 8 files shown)


tests/functional.rs
```diff
@@ -1732,7 +1732,7 @@ fn create_monero_syncer(
     let whost = env::var("MONERO_WALLET_HOST_2").unwrap_or("localhost".into());
     let lhost = env::var("MONERO_LWS_HOST").unwrap_or("localhost".into());
 
-    let lws_wallet = &format!("http://{}:38884", lhost);
+    let lws_wallet = &format!("http://{}:8443", lhost);
```
TheCharlatan, line 1735: I remapped this port from 8443 to 38884 intentionally to not use a port that is usually reserved for ssl web servers and to put it into a similar range as the other monero services.

h4sh3d: rollback that patch

PR #446, opened 2022-05-25
extract Swapd runtime state

zkao:


src/swapd/runtime.rs
```diff
@@ -1751,7 +1325,7 @@ impl Runtime {
                                 if self.state.a_refundsig()
                                     && !self.state.a_xmr_locked()
                                     && !self.state.a_buy_published()
-                                    && self.state.remote_params().is_some()
```
TheCharlatan, line 1754: Good catch!

TheCharlatan (approved):
ACK

PR #475, opened 2022-06-07
Peerd: Only send whitelisted Msg's over the bridge

TheCharlatan:
Only some of the messages should be allowed through the PeerdReceiverRuntime to the peerd Runtime. I don't like this explicit listing of messages, is there some other method to select this subset of enum variants?


h4sh3d:
There is the `matches!` macro in std but I'm not sure it's going to improve the code. I'm fine with the current patch

TheCharlatan:
> There is the `matches!` macro in std but I'm not sure it's going to improve the code. I'm fine with the current patch

I like this better, and pushed a patch.

h4sh3d (approved):

PR #476, opened 2022-06-07
Docs: Remove port reuse warning

TheCharlatan:


doc/install-guide.md
```diff
@@ -94,5 +94,3 @@ services:
     container_name: walletrpc
     command: "/usr/bin/monero-wallet-rpc --stagenet --disable-rpc-login --wallet-dir wallets --daemon-host stagenet.melo.tools:38081 --rpc-bind-ip 0.0.0.0 --rpc-bind-port 38083 --confirm-external-bind"
 ```
-
-:mag_right: If you plan to make offer with the Docker stack make sure you understand what ports need to be open and forwarded, a current limitation does not allow port reuse while making multiple offers. Either refresh the containers or forward more ports for peerd to listen on, by default `9735` is forwarded.
```
h4sh3d, line 98: The first part is still valid and a good hint. I think we can just remove from "a current limitation.."

TheCharlatan: Yeah, good point.

h4sh3d (approved):
ACK

PR #477, opened 2022-06-07
State Recovery management implementation

TheCharlatan:


doc/staterecovery_sequencediagram.txt
```diff
@@ -0,0 +1,32 @@
...
+== State Recovery Procedure
+entryspacing 0.8
+farcasterd -> checkpointd : launch
+checkpointd -> farcasterd : RecoverableStates
+farcasterd -> cli : RecoverableStates
+cli -> farcasterd : RecoverSelectedState
+farcasterd -> walletd : launch
+farcasterd -> swapd : launch
+farcasterd -> syncerd : launch
+farcasterd -> peerd : launch
```
zkao, line 21: here all the components are relaunched in a coupled manner, which probably doesnt reflect single component recovery, thus i guess if i ignore the lines related to other components and recover a single one, everything works, right?

src/databased/runtime.rs
```diff
@@ -225,6 +223,124 @@ impl Runtime {
     }
 }
 
+pub fn checkpoint_handle_multipart_receive(
+    checkpoint_multipart_chunk: request::CheckpointMultipartChunk,
+    pending_checkpoint_chunks: &mut HashMap<[u8; 20], HashSet<CheckpointChunk>>,
```
zkao, line 228: Why HashSet and not Vec?

src/farcasterd/runtime.rs
```diff
@@ -889,6 +895,33 @@ impl Runtime {
...
+                    endpoints.send_to(
+                        ServiceBus::Ctl,
+                        ServiceId::Farcasterd,
+                        source,
+                        Request::Failure(Failure {
+                            code: 1,
+                            info: "Cannot restore an already running swap".to_string(),
+                        }),
+                    )?;
+                    return Ok(());
```
zkao, line 909: this implies farcasterd Runtime will always have an up to date self.running_swaps

src/rpc/request.rs
```diff
@@ -715,10 +742,8 @@ pub struct Checkpoint {
 #[derive(Clone, Debug, Display, StrictDecode, StrictEncode)]
 #[display(Debug)]
 pub enum CheckpointState {
-    CheckpointWalletAlicePreLock(AliceState),
-    CheckpointWalletBobPreLock(BobState),
-    CheckpointWalletAlicePreBuy(AliceState),
-    CheckpointWalletBobPreBuy(BobState),
+    CheckpointWalletAlice(AliceState),
+    CheckpointWalletBob(BobState),
 }
```
zkao, line 747: An equivalent enum already exists as `Wallet` enum in wallet/runtime

src/walletd/runtime.rs
```diff
@@ -1533,6 +1545,30 @@ impl Runtime {
...
+                    if !self.wallets.contains_key(&swap_id) {
+                        self.wallets.insert(swap_id, Wallet::Alice(alice_state));
+                    }
+                }
+                CheckpointState::CheckpointWalletBob(bob_state) => {
+                    info!("Restoring bob wallet for swap {}", swap_id);
+                    if !self.wallets.contains_key(&swap_id) {
+                        self.wallets.insert(swap_id, Wallet::Bob(bob_state));
+                    }
+                }
```
zkao, line 1569: maybe add the else branches for logging errors?

src/rpc/request.rs
zkao: same for swapd/runtime, there it will be called State

src/rpc/request.rs
```diff
@@ -695,6 +703,25 @@ pub enum FundingInfo {
...
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_crate")
+)]
+#[display(NodeInfo::to_yaml_string)]
+pub struct CheckpointEntry {
+    pub swap_id: SwapId,
+    pub public_offer: PublicOffer<BtcXmr>,
+}
```
zkao, line 717: I didn't quite get why pub offer is in this struct?

zkao (commented):
Not a trivial PR to review, the logic related to the multipart message creates too much noise, and I tried to ignore those hacky bits, as they are temporary

I'd like to have some feedback on my comments, for the sake of understanding

src/rpc/request.rs
TheCharlatan: The checkpoint state is an amalgamation of state that extends beyond of what is currently encapsulated in the `Wallet` `State` enum, so I put them in separate structs. Since I could only test the restoration in #477 , the checkpoint struct are still undergoing some mutation.

src/databased/runtime.rs
zkao: How come they might not arrive on order?

zkao: again, this is part of multpart messages so no need to extend

src/rpc/request.rs
TheCharlatan: It's so users can see for which taken offers a swap was running. I might remove this again at a later stage, once we decide to either keep with the manual method, or automate everything and don't surface manual restoration to the user.

src/farcasterd/runtime.rs
zkao: Not sure. Maybe send a msg to swapd and if it doesn't error, then swapd is up and running, i guess?

src/walletd/runtime.rs
TheCharlatan: Done in 5f87e31ff2fa217c207f0eff83778932eb3f0fd5

src/rpc/request.rs
zkao: in farcasterd/Runtime there is consumed_offers field that tracks what offers were consumed

https://github.com/farcaster-project/farcaster-node/blob/4d6a27b5124be9755f8b02071ea2cfd65607040f/src/farcasterd/runtime.rs#L170

doc/staterecovery_sequencediagram.txt
```diff
@@ -1,32 +1,18 @@
...
-farcasterd -> checkpointd : RecoverSelectedState
-checkpointd -> walletd : RecoverWallet
-checkpointd -> peerd : RecoverPeerd
-checkpointd -> swapd : RecoverSwap
-// these tasks could be checkpointed, but can probably be recreated from scratch based on recovered state in swapd (and currently also walletd)
-swapd -> syncerd : WatchTasks
-swapd -> swapd : ResumeSwap
+farcasterd -> databased : launch
+cli -> farcasterd : RetrieveAllCheckpointInfo
+farcasterd -> databased : RetrieveAllCheckpointInfo
```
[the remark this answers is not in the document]
TheCharlatan: Good catch, this is totally wrong. Done in 0a1e4401001f3138a89c6ab448a9a4c46b93ad50

src/cli/command.rs
```diff
@@ -97,6 +97,16 @@ impl Exec for Command {
...
 
+            Command::ListCheckpoints => {
+                runtime.request(ServiceId::Database, Request::RetrieveAllCheckpointInfo)?;
+                runtime.report_response_or_fail()?;
+            }
+
+            Command::RestoreCheckpoint { swap_id } => {
+                runtime.request(ServiceId::Farcasterd, Request::RestoreCheckpoint(swap_id))?;
+                runtime.report_response_or_fail()?;
+            }
```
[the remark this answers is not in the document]
TheCharlatan: Yes, corrected in 0a1e4401001f3138a89c6ab448a9a4c46b93ad50

src/farcasterd/runtime.rs
TheCharlatan: It's pringing another ESB error if it send to a non-existing / dead endpoint. I'd like to avoid that.

zkao: is that true even if we handle the error at the level above internet2? meaning not letting error propagate to internet2

anyway possibly that is something that should end up in internet2 for checking if service is alive

TheCharlatan: Yeah, the error is logged at the ESB level first, before we can capture it in the runtime.

TheCharlatan: I removed the `running_swaps` check and am instead now checking if walletd is still running by sending a `Hello` request on the Msg bus.

Lederstrumpf: Good point @zkao. @TheCharlatan & I had a long discussion exploring multiple options for addressing this.
Ultimately, we should have a maintenance daemon that perform aliveness checks with all services. For now, we should just perform this check ad-hoc when restoring a checkpoint - regardless of *yet another ESB error*.

TheCharlatan: Checking if walletd is up in https://github.com/farcaster-project/farcaster-node/pull/477/commits/9734b0f93a0af4a3360a4598544fb43564d55cfe

Lederstrumpf (approved):

PR #482, opened 2022-06-08
feature: make progress waiting on new messages

h4sh3d:
This PR allows `progress` to pass a flag `-f|--follow` to continuously receive progress update until ctrl-c is pressed or termination condition is met: a request success or failure is received.

When the flag is not passed, the behavior is the same as previously, a one shot result in yaml format is returned.

#### How it works
When the client is in follow mode a first request is send to farcasterd to register itself in the progress subscription list, when a client register to a swap farcasterd send the queue back to the client. After registering the client loops on all messages until termination condition is met. When farcasterd receives a message related to some swap progress the message is forwarded to all client subscribed to it. If the forwarding fails the client is removed from the sub list.

```
| client |                | fcd |
    |                        |
   || ---------(sub)-------> || add to list
   ||                        ||
   || <--------------------- || send back queue
   || <--------------------- ||
   || <--------------------- ||
   ||                        |
   ||                        || <----(progress)
   || <--------------------- || call notify_subscribed_clients
   ||                        |
   ||                        || <----(success or failure)
   || <--------------------- || call notify_subscribed_clients
   ||                        |
   || --------(unsub)------> || remove from list
    |                        |
```
If the client has ctrl-c before unsubscribing `call notify_subscribed_clients` will remove it from the list if send fails.

#### This PR does not ensure the progress messages are send in the correct order; this is out of the scope here. Every swap progress sequence should match the following request pattern:
```
 ( Progress::Message | Progress::StateTransition )*
 Success | Failure
```
Subsequent messages will be ignore is some cases (but not when displayed through yaml format).

### DIFF
--- a/src/rpc/client.rs
+++ b/src/rpc/client.rs
@@ -102,7 +102,7 @@ impl Client {
 
     /// Print the stream of received requests until progress fails or succeed
     pub fn report_progress(&mut self) -> Result<(), Error> {
-        // loop on all requests received until a progress terminaison condition is recieved
+        // loop on all requests received until a progress termination condition is recieved
         // report failure transform Request::Failure in error already, terminate on error or on
         // success
         let res = loop {
--- a/src/rpc/client.rs
+++ b/src/rpc/client.rs
@@ -14,13 +14,14 @@
 
 use crate::service::Endpoints;
 use std::convert::TryInto;
-use std::thread::sleep;
+use std::thread::{self, sleep};
 use std::time::Duration;
 
 use internet2::ZmqType;
 use microservices::esb;
 
 use crate::rpc::request::OptionDetails;
+use crate::rpc::request::SwapProgress;
 use crate::rpc::{Request, ServiceBus};
 use crate::service::ServiceConfig;
 use crate::{Error, LogStyle, ServiceId};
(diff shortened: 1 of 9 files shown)


src/rpc/request.rs
```diff
@@ -525,23 +525,35 @@ pub enum Request {
...
+    #[display(inner)]
     SwapProgress(SwapProgress),
 
     #[api(type = 1003)]
     #[display("read_progress({0})")]
     ReadProgress(SwapId),
 
+    #[api(type = 9999)]
+    #[display("test_progress({0})")]
+    TestProgress(SwapId),
```
TheCharlatan, line 541: Unused

src/rpc/client.rs
```diff
@@ -92,38 +93,33 @@ impl Client {
...
-            finished = true;
-            counter += 1;
-            match self.report_failure()? {
-                // Failure is already covered by `report_response()`
-                Request::Progress(info) => {
-                    println!("{}", info.green_bold());
-                    finished = false;
+    /// Print the stream of received requests until progress fails or succeed
+    pub fn report_progress(&mut self) -> Result<(), Error> {
+        // loop on all requests received until a progress terminaison condition is recieved
```
TheCharlatan, line 105: typo s/terminaison/termination/

src/rpc/client.rs
```diff
@@ -92,38 +93,33 @@ impl Client {
...
-                Request::Success(OptionDetails(Some(info))) => {
-                    println!("{}{}", "Success: ".bright_green_bold(), info);
-                }
-                Request::Success(OptionDetails(None)) => {
-                    println!("{}", "Success".bright_green_bold());
-                }
-                other => {
-                    eprintln!("{}: {}", "Unexpected report".err(), other.err_details());
-                    return Err(Error::Other(s!("Unexpected server response")));
+                Ok(Request::Success(s)) => {
```
TheCharlatan, line 114: We should open an issue to add a Success message once the swap has completed.

TheCharlatan (changes requested):
Nice work, just two tiny nits. I'm sure we can re-use this to stream progress to other APIs.

src/rpc/client.rs
h4sh3d: Done in be8d233d9545c3143bdc3030b7587b21cade6952

src/rpc/request.rs
h4sh3d: Fixed in 109f6125744e7ac9a651cc08cb7f78d82606c37c

TheCharlatan (approved):

PR #484, opened 2022-06-08
Farcasterd: Check if peerd listen launched successfully

TheCharlatan:
Uses the same heuristic as the `connect_peer` method.


h4sh3d (approved):
tACK with added changes. Fix when peerd was registered in listens before crash.

TheCharlatan (commented):
ACK on your changes.

PR #489, opened 2022-06-10
feature: reorganize and split tests

h4sh3d:
Split tests into more files, each test start with `{file}_` e.g. `swap_` in `swap.rs` or `monero_` in `monero.rs`. Test are run the same way as before `cargo tests {monero|bitcoin|swap|...} ...`.

List of tests category/files:
- swap
- monero
- bitcoin
- grpc
- cli

`utils` module should still receive more code from root test files and should later be improved to help stabilizing the tests.

#### This PR aim to not alter tests behavior, just to reorganize so we can work on test improvments independently without conflict later


TheCharlatan (changes requested):

.github/workflows/test.yml
```diff
@@ -53,9 +53,25 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test: [bitcoin, monero, swap]
```
TheCharlatan, line 56: The timeout-minutes should be significantly reduced again. I think we can go back to the old 45 min limit.

h4sh3d: Done 9ecacfbae4a48c5dd20c544274f9b7849c117eab

tests/utils/mod.rs
```diff
@@ -1,55 +1,14 @@
...
-    let node = NodeConfig {
-        host: "localhost".to_string(),
-        port: 123,
-        protocol: None,
-    };
-    let display = node.to_string();
-    assert_eq!(display, "localhost:123");
+pub fn setup_logging() {
+    // !!! Configure RUST_LOG in CI to change this value !!!
+    let env = env_logger::Env::new().default_filter_or("farcaster_node=info");
```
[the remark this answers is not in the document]
h4sh3d: Done 9ecacfbae4a48c5dd20c544274f9b7849c117eab

h4sh3d:
I want to fix and upgrade grpc in another PR. Looks like binding on ipv6 might be the problem, but we currently cannot change through config. If ok we can merge it as is

TheCharlatan:
I added a commit with a TODO to disable grpc tests for now.

TheCharlatan (approved):

PR #496, opened 2022-06-16
Cli: Add endpoint to print information about a passed-in offer

TheCharlatan:
This completes #457 by adding an endpoint to display information about an offer. It is similar to the `take` command but does not interact with any of the other microservices.


zkao (approved):

h4sh3d (approved):
utACK

PR #497, opened 2022-06-20
State recovery2 rebased

zkao:


zkao:
this is a simple rebase. the intention is to replace #477 (and avoid messing up with that PR)

PR #506, opened 2022-06-23
Persist funding address secret key

TheCharlatan:
Add endpoints to the database to persist addresses and their private keys. Does not include actual integration with the rest of the codebase, e.g. persisting the addresses when the funding keys are created.

### DIFF
--- a/src/databased/runtime.rs
+++ b/src/databased/runtime.rs
@@ -494,6 +494,11 @@ impl Database {
                 &secret_key.secret_bytes(),
                 lmdb::WriteFlags::empty(),
             )?;
+        } else {
+            warn!(
+                "address {} was already persisted with its secret key",
+                address
+            );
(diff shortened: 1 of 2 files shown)


src/databased/runtime.rs
```diff
@@ -429,17 +464,72 @@ impl From<Vec<u8>> for CheckpointKey {
...
+        let mut key = vec![];
+        let _key_size = address.strict_encode(&mut key);
+        if tx.get(db, &key).is_err() {
+            tx.put(
+                db,
+                &key,
+                &secret_key.secret_bytes(),
+                lmdb::WriteFlags::empty(),
+            )?;
+        }
```
zkao, line 497: on else log key exists

zkao (approved):
excellent, clean and simple

PR #507, opened 2022-06-25
Persist offer history

TheCharlatan:
Adds persistance for offers. Offers can now be written to a separate table in the lmdb database.

They can be retrieved by selecting their status. The `OfferStatus` and `OfferStatusSelector` are separated to allow for extension of their criteria. For example saving addresses in the `Open` status, or `SwapId`s in the `InProgress` status, and selecting offers based on other common criteria, like amounts.

### DIFF
--- a/src/databased/runtime.rs
+++ b/src/databased/runtime.rs
@@ -1,3 +1,6 @@
+use crate::databased::runtime::request::OfferStatus;
+use crate::databased::runtime::request::OfferStatusPair;
+use crate::databased::runtime::request::OfferStatusSelector;
 use crate::farcaster_core::consensus::Encodable;
 use crate::walletd::runtime::{CheckpointWallet, Wallet};
 use farcaster_core::negotiation::PublicOffer;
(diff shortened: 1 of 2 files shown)


src/rpc/request.rs
```diff
@@ -704,9 +704,66 @@ pub enum Request {
...
+)]
+pub enum OfferStatus {
+    #[display("Open")]
+    Open,
+    #[display("In Progress")]
+    InProgress,
+    #[display("Ended({0})")]
+    Ended(Outcome),
+}
+
```
zkao, line 760: `Expired`?

zkao (approved):
lgtm, minor comments only

src/databased/runtime.rs
```diff
@@ -466,18 +483,71 @@ struct Database(lmdb::Environment);
 
 const LMDB_CHECKPOINTS: &str = "checkpoints";
 const LMDB_ADDRESSES: &str = "addresses";
+const LMDB_OFFER_HISTORY: &str = "history";
```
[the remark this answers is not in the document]
TheCharlatan: Ah, good point.

src/rpc/request.rs
TheCharlatan: `Expired` has its own semantics for me, I'd rather think of it as a variant of the `Outcome` enum.

src/databased/runtime.rs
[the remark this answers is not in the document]
TheCharlatan: Done in 02aee9184fa675c5d9258858e005d41cb8d4377c

src/rpc/request.rs
zkao: I see. For me `Outcome` is about a swap, not an offer, and an offer that was never taken never becomes an `Outcome`, its just an expired offer

zkao: For me a `Trade` is a swap, not an `Offer`, and only a trade has an `Outcome`, so I don't see how that helps

zkao: But it doesnt matter for now anyway

h4sh3d: `Trade` would be different than `Offer`: this should allow use to have a proper `Offer` type (with room for negotiation) and a `Trade` type (close to what we currently have in `Offer`) that represent all the parameters of a swap.

`Offer` is kind of in between currently.

PR #514, opened 2022-06-30
shell.nix: update dev environmnet

zkao:


TheCharlatan (approved):
I'm sure this is fine.

PR #518, opened 2022-07-02
Swap: Allow cancel for alice refundsig before arb lock

TheCharlatan:
Alice should be able to cancel the swap up until the point where Bob's lock has confirmed.


zkao (approved):
simple, and add btc locked state that was previously missing

PR #520, opened 2022-07-02
Swap: Ensure a checkpoint is only set once and in the correct order

TheCharlatan:
This also solves the problem where a checkpoint of the same state would be written to the database again on restore.

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -38,7 +38,7 @@ use std::{
 
 use super::{
     storage::{self, Driver},
-    swap_state::{AliceState, BobState, State},
+    swap_state::{AliceState, BobState, State, SwapCheckpointType},
     syncer_client::{log_tx_received, log_tx_seen, SyncerState, SyncerTasks},
     temporal_safety::TemporalSafety,
 };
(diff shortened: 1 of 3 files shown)


src/swapd/swap_state.rs
```diff
@@ -481,4 +515,60 @@ impl State {
...
+        if let State::Bob(BobState::RevealB { checkpoint, .. }) = self {
+            if checkpoint.is_none() {
+                *checkpoint = Some(SwapCheckpointType::CheckpointBobPreLock);
+                true
+            } else {
+                debug!("checkpoint bob pre lock already set");
+                false
+            }
+        } else {
+            error!("Not on CoreArbB state");
```
TheCharlatan, line 556: Wrong message.

src/swapd/swap_state.rs
```diff
@@ -75,14 +78,18 @@ pub enum BobState {
...
         cancel_seen: bool,
         remote_params: Params,
         local_params: Params,
         b_address: bitcoin::Address,
     }, // lock (not signed), cancel_seen, remote
     #[display("BuySig")]
-    BuySigB { buy_tx_seen: bool },
+    BuySigB {
+        buy_tx_seen: bool,
+        checkpoint: SwapCheckpointType,
```
zkao, line 91: `checkpoint_type`?

zkao: when i read, I assumed the entire checkpoint was in the state

src/swapd/runtime.rs
```diff
@@ -771,22 +771,24 @@ impl Runtime {
...
-                                last_msg: Msg::BuyProcedureSignature(buy_proc_sig.clone()),
-                                enquirer: self.enquirer.clone(),
-                                temporal_safety: self.temporal_safety.clone(),
-                                txs: self.txs.clone(),
-                                txids: self.syncer_state.tasks.txids.clone(),
-                                pending_requests: self.pending_requests.clone(),
-                                xmr_addr_addendum: self.syncer_state.xmr_addr_addendum.clone(),
-                            }),
-                        )?;
+                        if self.state.a_sup_checkpoint_pre_buy() {
```
zkao, line 774: this one comment extends to the rest of the pr related to state updates, sup. Here one updates the state before being sure that checkpoint_send is successful, therefore generating some potential invalid and unhandled cases

zkao (approved):
Excellent PR, minor comments related to QA

src/swapd/swap_state.rs
TheCharlatan: `last_checkpoint_type` :)

TheCharlatan: Done in f44aa830eabd6de683354f1cb197b45287943fe3

src/swapd/runtime.rs
TheCharlatan: Yeah, we need a workflow for critical calls like this where we wait for a response from the database before progressing in the protocol. But I need to think about this some more, so work for another PR.

zkao: you're right, thats a larger problem

Issue #515, opened 2022-06-30
Race condition between broadcast lock and bob pre buy checkpoint


TheCharlatan:
There is a race condition between bob cehckpointing pre buy and the lock transaction being broadcast. The syncer might not have broadcasted the transaction yet, when the checkpoint is already set. If the swap is shutdown at that point in time, it won't recover correctly, since it assumes the lock transaction is already broadcasted. It won't attempt re-broadcast either.

PR #521, opened 2022-07-02
Fix Checkpoint race condition

TheCharlatan:
Solves the race condition described in #515. The syncer state (syncer client) now keeps track of transactions that are pending broadcast. Once their broadcast has been confirmed by the syncer, they are removed again.

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -153,6 +153,7 @@ pub fn run(
         watched_txs: none!(),
         retrieving_txs: none!(),
         sweeping_addr: none!(),
+        broadcasting_txs: none!(),
         txids: none!(),
         final_txs: none!(),
         tasks: none!(),
@@ -262,6 +263,7 @@ pub struct CheckpointSwapd {
     pub temporal_safety: TemporalSafety,
     pub txs: HashMap<TxLabel, bitcoin::Transaction>,
     pub txids: HashMap<TxLabel, Txid>,
+    pub pending_broadcasts: Vec<bitcoin::Transaction>,
     pub pending_requests: HashMap<ServiceId, Vec<PendingRequest>>,
 }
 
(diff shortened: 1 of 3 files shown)


src/swapd/syncer_client.rs
```diff
@@ -283,6 +288,38 @@ impl SyncerState {
...
         task
     }
+    pub fn broadcast(&mut self, tx: bitcoin::Transaction) -> Task {
+        let id = self.tasks.new_taskid();
+        let task = Task::BroadcastTransaction(BroadcastTransaction {
+            id,
+            tx: bitcoin::consensus::serialize(&tx),
+        });
+        self.tasks.tasks.insert(id, task.clone());
+        self.tasks.broadcasting_txs.insert(id);
```
zkao, line 298: I'm afraid there is redundancy here

tasks.tasks keeps track of ongoing tasks, and if the task is finished, the associated data should be removed

it looks like broadcasting_txs does exactly the same

zkao: oh, i misread, its just the task id that is being stored, sorry for the confusion

src/swapd/runtime.rs
```diff
@@ -262,6 +263,7 @@ pub struct CheckpointSwapd {
     pub temporal_safety: TemporalSafety,
     pub txs: HashMap<TxLabel, bitcoin::Transaction>,
     pub txids: HashMap<TxLabel, Txid>,
+    pub pending_broadcasts: Vec<bitcoin::Transaction>,
```
zkao, line 266: I'm not convinced this is needed since that data is in tasks.tasks

TheCharlatan: I prefer it this way, since it is more transparent on what the actual contents of the checkpoint are, as opposed to the more opaque `Task`. Besides, you'll have to re-assemble the task in any case on checkpoint restore, since it will need a new lifetime and id.

zkao (approved):

PR #523, opened 2022-07-02
Naming: Rename cancel swap to abort swap

### LINKED ISSUE #516: Terminology: Cancel swap vs Abort swap
we were using cancel for the cancel tx in the refund/punish path

now we are using it as well for manually aborting/cancelling the swap, which does not necessarly involve the cancel tx.

i find this to be confusing for a newcomer

TheCharlatan:
Closes #516


zkao (approved):
lgtm, pure renaming

TheCharlatan:
Rebased.

PR #524, opened 2022-07-03
Swap: Add public offer to GetInfo

### LINKED ISSUE #458: Add a mapping from SwapId to Offer information
Towards fulfilling the requirements of #258 it should be possible to retrieve information that was originally part of the offer (e.g. exchange rate, counterparty, thresholds, etc.) from a running swap.

TheCharlatan:
Adds a public_offer to each swap state. Not sure if the swap state is the correct place to put it, as opposed to the Runtime, would be great to get your comment on this @zkao.

Also cleans up the swap GetInfo struct a bit.

Closes #458

### DIFF
--- a/src/rpc/request.rs
+++ b/src/rpc/request.rs
@@ -1177,19 +1177,11 @@ pub struct SwapInfo {
     #[serde_as(as = "Option<DisplayFromStr>")]
     pub swap_id: Option<SwapId>,
     // pub state: crate::swapd::State,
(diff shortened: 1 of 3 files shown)


zkao:
i would put this straight up on the runtime, because its not state machine related

TheCharlatan:
> i would put this straight up on the runtime, because its not state machine related
Done.

zkao (approved):
lgtm

PR #526, opened 2022-07-04
swapd: add tasks to CheckpointSwapd

zkao:


TheCharlatan:
NACK on this, this seems to skip of the logic that gets triggered when the tasks are submitted piecemeal instead. The logic could be re-implemented, but this would just be a duplication.

PR #533, opened 2022-07-04
swapd: fee estimate from syncer (electrum)

zkao:


src/swapd/runtime.rs
```diff
@@ -544,6 +554,10 @@ impl Runtime {
                             self.syncer_state.monero_syncer(),
                             Request::SyncerTask(watch_height_xmr_task),
                         )?;
+
+                        // syncer takes too long to give a fee
+                        std::thread::sleep(Duration::from_secs_f32(5.0));
+
```
zkao, line 560: @TheCharlatan, the fee_estimate command takes far too long to return from Syncer to Swapd.

Could we poll and cache values at the level of the syncer, and those are returned directly. i'd say every 20 s should be enough.

zkao: The sleep there is to avoid the race. 

A lot of the things happen, message goes through the remote peer, and comes back and we still dont have a fee estimate

TheCharlatan: The caching can be done, yes. Nevertheless it would be nice to not rely on the fee estimator to return in time. Do we make similar assumptions for other requests?

zkao:
The syncer is the one that is interacting externally, so we can't rely messages will be delivered in time. There are potential race conditions there, since we use Events to trigger actions in swapd (the sequence diagrams is the place to search for them). Especially at launch when too many messages are flying around.

As you probably know from the sequence diagrams, we do rely that for a given service bus, its msgs are received in the order they are sent.

And if there will be a race condition because one msg is being sent through Msg bus and the other through Ctl bus, then we have to put the two in the same bus, so that they're ordered correctly, and thus received in the correct order

------- Original Message -------

zkao:
Most of the tasks are `watch` tasks. So those are easy to handle, because u subscribe very well ahead of time.

While coding, one simplification that I did and that might still be around is, for example, that the act of sending a broadcast task actually broadcasts a tx. Haven't searched for those.

src/swapd/runtime.rs
TheCharlatan: Ok, I'm reworking the entire fee estimation task for the syncer now.

src/syncerd/types.rs
```diff
@@ -305,10 +305,11 @@ pub struct FeeEstimation {
 
 #[derive(Clone, Debug, Display, StrictEncode, StrictDecode, Eq, PartialEq, Hash)]
 #[display(Debug)]
+// the sats per kvB is because we need u64 for Eq, PartialEq and Hash
```
TheCharlatan, line 308: Thank you for the comments and I like the increased precission we get from calculating per kvbyte instead of vbyte.

src/syncerd/bitcoin_syncer.rs
```diff
@@ -1124,3 +1114,18 @@ fn logging(txs: &[AddressTx], address: &BtcAddressAddendum) {
         );
     });
 }
+
+/// Input fee in sat_per_kvb, output fee in sat units
+pub fn signed_tx_fee(fee_sat_per_kvb: u64, unsigned_tx_vsize: usize, nr_inputs: usize) -> u64 {
```
TheCharlatan, line 1119: Maybe to make it clear, rename this to `p2wpkh_signed_tx_fee`.

TheCharlatan:
From the logs:

```
[2022-07-11T08:38:45Z INFO  farcaster_node::swapd::runtime] 0xe2ad…44ae | Accepting swap as Maker from Taker through peerd peerd<03ee5e61d42bc0f932c0e7d09e0df6ab8b2dde43d86131645cd4b074a738a56803@0.0.0.0:9376>
[2022-07-11T08:38:45Z DEBUG microservices::esb::controller] swap<0xe2ad…44ae> -> farcasterd: progress: Accepting swap 0xe2ad…44ae as Maker from Taker through peerd peerd<03ee5e61d42bc0f932c0e7d09e0df6ab8b2dde43d86131645cd4b074a738a56803@0.0.0.0:9376>
[2022-07-11T08:38:45Z ERROR microservices::esb::controller] ESB request processing error: error sending message from swap<0xe2ad…44ae> to Bitcoin (Local) syncer. Details: service is offline or not responding
[2022-07-11T08:38:45Z DEBUG syncerd] MSG RPC socket ipc://tests/fc1/msg.rpc
[2022-07-11T08:38:45Z DEBUG syncerd] CTL RPC socket ipc://tests/fc1/ctl.rpc
[2022-07-11T08:38:45Z DEBUG syncerd] Starting runtime ...
```

Seems like the problem is when Bob is Maker, the swap daemon receives `MakeSwap` before the syncers are up. Either this requires another sleep before the EstimateFee Task is spawned, or we wait for some kind of `Hello` from the syncer before spawning syncer tasks in the swap daemon, or we spawn the syncers once the offers are open instead of when the swap is initialized.

TheCharlatan:
I think you fixed it for MakerCommit (Bob Taker), but the problem I was referring to is triggered when Bob is Maker.

doc/sequencediagram.txt
```diff
@@ -56,11 +56,13 @@ t_swap <- peerd : Msg MakerCommit
 // TODO: verify that swapd launches no matter what
 t_syncer <- t_swap : Ctl WatchHeight
 t_syncer <- t_swap : if Bob, Watch Arbitrating Funding Address
+t_syncer<-t_swap:If Bob, Ctl EstimateFee (btc)
```
TheCharlatan, line 59: Can you commit the new `.svg` as well?

src/swapd/runtime.rs
```diff
@@ -1002,6 +1044,18 @@ impl Runtime {
...
+                let btc_fee_task = self.syncer_state.estimate_fee_btc();
+                endpoints.send_to(
+                    ServiceBus::Ctl,
+                    self.identity(),
+                    self.syncer_state.bitcoin_syncer(),
+                    Request::SyncerTask(btc_fee_task),
+                )?;
+
+                // syncer takes too long to give a fee
+                std::thread::sleep(Duration::from_secs_f32(10.0));
```
TheCharlatan, line 1057: I've run the PR successfully for 10 times with 0.5 seconds delay instead here. I think it is safe enough to at least halve it.

TheCharlatan (changes requested):

src/swapd/runtime.rs
```diff
@@ -642,15 +663,23 @@ impl Runtime {
...
 
-                                if let Some(address) = self.state.b_address().cloned() {
+                                if let (Some(address), Some(sat_per_kvb)) = (
+                                    self.state.b_address().cloned(),
+                                    self.syncer_state.btc_fee_estimate_sat_per_kvb,
+                                ) {
                                     let swap_id = self.swap_id();
-                                    let fees = bitcoin::Amount::from_sat(200); // FIXME
-                                    let amount = self.syncer_state.bitcoin_amount + fees;
+                                    let vsize = 94;
```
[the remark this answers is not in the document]
TheCharlatan: Fair enough :)

src/swapd/runtime.rs
TheCharlatan: The `swap_bob_maker_user_abort_sweep_btc` functional test aborts right after funding, so it is perfect to test this fast.

TheCharlatan:
The parallel swap test is failing now, no idea why, upon investigation it seems like 0.5 seconds are too fast for it. I bumped it to 2.0 and it is working again.

TheCharlatan:
> ```
> 2022-07-11T12:46:55.3239843Z [2022-07-11T12:46:55Z ERROR farcaster_node::swapd::runtime] Incorrect amount funded. Required: 1.00000124 BTC, Funded: 1.00000123 BTC. Do not fund this swap anymore, will abort and atttempt to sweep the Bitcoin to the provided address.
> 2022-07-11T12:58:13.9790261Z [2022-07-11T12:58:13Z DEBUG microservices::esb::controller] swap<0x575d…5dc0> -> farcasterd: progress: Too big amount funded. Required: 1.000000000000 XMR, Funded: 1.000000000001 XMR. Do not fund this swap anymore, will attempt to refund.
> ```

@zkao , from which tests are these logs? Aren't they from the overfunding tests and therefore expected and correct behavior?
There are two overfunding tests, one for Alice, one for Bob, so in the entire log, you should see exactly these two occurences once.

The Bob incorrect funding test is here:  https://github.com/farcaster-project/farcaster-node/blob/main/tests/functional-swap.rs#L75
The Alice overfunding funding test is here: https://github.com/farcaster-project/farcaster-node/blob/main/tests/functional-swap.rs#L259

Please state which test is failing because of these errors before proceeding.

> i understand u guys dont want epsilon, but then how do i ensure all amounts are always exact at the sat or atomic unit level? reference from this discussion: https://github.com/farcaster-project/farcaster-node/pull/534

In which scenario is the funding amount not exact right now?

TheCharlatan (approved):
Thank you for bearing through this one, I am happy we are this far now :)

PR #561, opened 2022-07-12
Swapd: create better API for handling PendingRequests

zkao:


src/swapd/runtime.rs
```diff
@@ -2258,8 +2337,9 @@ impl Runtime {
                 }
                 // set external eddress: needed to subscribe for buy tx (bob) or refund (alice)
                 self.syncer_state.tasks.txids.insert(TxLabel::Buy, txid);
-
+                // self.defer_request()
```
TheCharlatan, line 2340: Remove

src/swapd/runtime.rs
```diff
@@ -445,6 +468,52 @@ impl Runtime {
...
+                            _ => endpoints
+                                .send_to(
+                                    r.bus_id.clone(),
+                                    r.source.clone(),
+                                    r.dest.clone(),
+                                    r.request.clone(),
+                                )
+                                .map_err(Into::into),
+                        } {
+                            success = success && true;
```
TheCharlatan, line 499: I don't understand this line. Success starts out as true. In this line, it is set to true, if it was true previously, and false if it was already false.

src/swapd/runtime.rs
```diff
@@ -445,6 +468,52 @@ impl Runtime {
...
+                            None
+                        } else {
+                            success = false;
```
TheCharlatan, line 502: Log the error?

src/swapd/runtime.rs
```diff
@@ -445,6 +468,52 @@ impl Runtime {
...
+                            Some(r)
+                        }
+                    } else {
+                        Some(r)
+                    }
+                })
+                .collect();
+            self.pending_requests.insert(key, remaining_pending_reqs);
+        } else {
+            success = false;
```
TheCharlatan, line 512: Again log an error?

src/swapd/runtime.rs
```diff
@@ -445,6 +468,52 @@ impl Runtime {
         }
     }
 
+    fn continue_deferred_requests(
+        &mut self,
+        endpoints: &mut Endpoints,
+        key: ServiceId,
+        predicate: fn(&PendingRequest) -> bool,
+    ) -> bool {
```
TheCharlatan, line 476: What is the vision behind returning this bool?

TheCharlatan (changes requested):

PR #567, opened 2022-07-13
Prepare update when core 0.5.0 is released

h4sh3d:
This will not compile, it points to local repo on filesystem and doesn't compile yet. I open it for visibility only.


TheCharlatan:
These are far fewer changes than I would have expected :)

TheCharlatan (approved):
ACK

PR #569, opened 2022-07-13
swapd: create wallet client (for better logs)

zkao:


TheCharlatan:
This is just moving a small of logging code. What is your vision for the wallet client and what benefit should it provide?

TheCharlatan:
Then why not move it to swapd? Seems confusing to log as wallet client if there is no such thing used for anything else.

PR #583, opened 2022-07-22
Farcasterd syncer types

zkao:


src/farcasterd/runtime.rs
```diff
@@ -2233,3 +2251,83 @@ pub fn launch(
         err
     })
 }
+#[derive(Default)]
+struct Syncers(HashMap<(Coin, Network), Syncer>);
```
h4sh3d, line 2255: This should be `type Syncers = HashMap<...>` instead I think.

TheCharlatan:
Can you rebase this so we can run the improved CI against it?

src/farcasterd/runtime.rs
h4sh3d: Either you go back and implement some Deref trait or similar to avoid forwarding, or you keep the type definition and use a trait to implement custom methods on the HashMap, e.g

```rust
type Link = Option<Box<Node>>;

trait LinkMethods {
    fn cool_method(&self);
}

impl LinkMethods for Link {
    fn cool_method(&self) {
        // ...
    }
}
```

PR #588, opened 2022-07-24
use syncer types and associated fns to avoid race condition between swapd, monero syncer and bitcoin syncer

zkao:


src/farcasterd/runtime.rs
```diff
@@ -2233,3 +2266,68 @@ pub fn launch(
...
+
+impl Syncers {
+    pub fn inner(&mut self) -> &mut HashMap<(Coin, Network), Syncer> {
+        &mut self.0
+    }
+    pub fn syncer_services_len(&self) -> usize {
+        self.0
+            .values()
+            .filter(|Syncer { service_id, .. }| service_id.is_some())
+            .collect::<Vec<_>>()
```
TheCharlatan, line 2280: Use `.count()` instead

src/farcasterd/runtime.rs
```diff
@@ -715,7 +740,10 @@ impl Runtime {
                     } else {
                         let xs = self
                             .pending_swap_init
-                            .entry(ServiceId::Syncer(Coin::Bitcoin, *network))
+                            .entry((
+                                ServiceId::Syncer(Coin::Bitcoin, *network),
+                                ServiceId::Syncer(Coin::Monero, *network),
+                            ))
```
TheCharlatan, line 746: These two if/else cases are super hard to read with all the code duplication going on. Not the problem of this PR, but something I found distracting.

TheCharlatan:
> to be merged after https://github.com/farcaster-project/farcaster-node/pull/583

I found this PR much easier to review by itself seeing that there was no dead code and I could actually understand what the intention behind certain designs is. There is just one small change I suggested, then I think this can be merged.

TheCharlatan (approved):
ACK
@zkao I will let your merge.

PR #589, opened 2022-07-25
Bump internet2 and microservices to v0.8

TheCharlatan:
This PR does extensive changes to the node by removing the lightning dependencies (which became annoying to keep with the new microservices api definitons) and switching to the new internet2 address format.

internet2-address-types

With internet2 dropping most of the lightning-specific behavior, we could get rid of things like the default lightning port and lightning compatibility messages like `Ping` and `Pong`. This allowed me to remove the lightning dependencies entirely.

The minimum supported rust version had to be bumped because internet2 0.8 requires it. 

I also got rid of the reply and message file that we practically did not use. This mandated however that we define our own `Failure` type, which was another bi-product of navigating around an otherwise complicated change introduced by `microservices 0.8`. I added a TODO to add further failure codes eventually.


src/farcasterd/runtime.rs
```diff
@@ -993,7 +988,7 @@ impl Runtime {
                     self.listens
                         .clone()
                         .values()
-                        .map(|listen| listen.to_url_string()),
+                        .map(|listen| format!("{}", listen)),
```
h4sh3d, line 991: Probably works with `.to_string()`

src/lib.rs
```diff
@@ -30,8 +30,6 @@ extern crate farcaster_core;
 extern crate amplify;
 #[macro_use]
 extern crate amplify_derive;
-#[macro_use]
-extern crate lnpbp;
```
h4sh3d, line 34: NICE ! :tada:

h4sh3d (approved):

src/farcasterd/runtime.rs
TheCharlatan: Done c8efb6b5d5b809232544f7477dcaedc403c446e1

PR #590, opened 2022-07-25
Use `RUST_LOG` to configure loggers instead of `--verbose` flag

h4sh3d:
To allow more granular settings in logger this PR switch from using one flag `--verbose` and instead make use of the `RUST_LOG` environment variable.

By default, if the env var is not set, the default is `farcaster_node=info`. This will display all info logs ONLY coming from farcaster node. In test the var is set to `farcaster_node=debug,microservices=debug`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,7 +135,7 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2.0.0
         with:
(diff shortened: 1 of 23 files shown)


h4sh3d:
@TheCharlatan I'm done with this one, feel free to review&merge

TheCharlatan (approved):
Thank you for working on this

PR #592, opened 2022-07-26
Database: Log debug, not error, when removed entry does not exist

TheCharlatan:
Checkpoint removal used to log an `Error` every time a swap was cleaned up and no checkpoint data for the swap had been written yet. This PR relegates this database log to `Debug`.

### DIFF
--- a/src/databased/runtime.rs
+++ b/src/databased/runtime.rs
@@ -268,14 +268,18 @@ impl Runtime {
             }
 
             Request::RemoveCheckpoint(swap_id) => {
-                self.database.delete_checkpoint_state(CheckpointKey {
+                if let Err(err) = self.database.delete_checkpoint_state(CheckpointKey {
                     swap_id,
                     service_id: ServiceId::Wallet,
-                })?;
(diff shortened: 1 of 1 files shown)


src/databased/runtime.rs
```diff
@@ -268,14 +268,18 @@ impl Runtime {
...
 
             Request::RemoveCheckpoint(swap_id) => {
-                self.database.delete_checkpoint_state(CheckpointKey {
+                if let Err(err) = self.database.delete_checkpoint_state(CheckpointKey {
                     swap_id,
                     service_id: ServiceId::Wallet,
-                })?;
-                self.database.delete_checkpoint_state(CheckpointKey {
+                }) {
+                    debug!("{} | Did not delete checkpoint entry: {}", swap_id, err);
```
h4sh3d, line 275: Maybe we could use `warn!` here?

TheCharlatan: This happens too often, no need to warn. It's also just for deletion, so it is not that important.

h4sh3d (approved):
@TheCharlatan I let you merge if you're done with it!

PR #595, opened 2022-07-26
Checkpointing: Remove multipart message chunking

TheCharlatan:
We now have unlimited size zmq messages, so remove the checkpoint multipart chunking. :tada: :smile:


h4sh3d (approved):
ACK

PR #623, opened 2022-08-02
Cli: Print both serialized and deserialized offers in list-offers

TheCharlatan:


h4sh3d:
CI fails because of lws, after #627 get merged rebase and it should be fine again.

PR #614, opened 2022-08-01
ci: bumps containers in tests

h4sh3d:
Use latest Bitcoin, Electrs, and Monero versions

PR #626, opened 2022-08-02
ci: update monerod and monero-wallet-rpc versions

LeoNero:


h4sh3d:
I'm fine bumping these before #614, but check also for lws bump. I think ci will fail anyway.

Can you deactivate the lws test that fails in monero_syncer tests please?

h4sh3d:
#627 should go first to fix the current versions in CI, then rebase and replace with your bump.

LeoNero:
@h4sh3d should just wait for #614 or should still deactivate test on monero_syncer?

h4sh3d:
> @h4sh3d should just wait for #614 or should still deactivate test on monero_syncer?

lws "should" work with tag `monerod-0.17.3.2` (what why I did #614), but doesn't work with 0.18 stack (at least not when I tried), so yes you can bump to 0.18 and deactivate lws so we can use the 0.18 without lws.

LeoNero:
@h4sh3d done

LeoNero:
@h4sh3d test ignored

LeoNero:
Close since #614 already updates `monerod` and `monero-wallet-rpc`

PR #637, opened 2022-08-08
swapd: improve PendingRequests and add associated fn defer_request

zkao:


src/swapd/runtime.rs
```diff
@@ -210,16 +210,71 @@ pub struct Runtime {
...
+        } else {
+            error!("no request consumed with this predicate");
+            false
+        };
+        success
     }
 }
 
-impl PendingRequests {}
+trait PendingRequestsT {
```
TheCharlatan, line 269: What is the purpose of this trait? Is this interface reused somewhere else?

TheCharlatan: Go it!

h4sh3d: Common pattern to extend features from foreign types, the most often I saw it as `WriteExt: std::io::Write` to extend writer capabilities (e.g. in encoding/decoding rust-bitcoin).

PR #638, opened 2022-08-08
swap state: improve readability

zkao:


src/swapd/swap_state.rs
```diff
@@ -520,7 +520,8 @@ impl State {
                 &*self
             );
             return false;
-        } else if self.b_buy_tx_seen() {
+        };
```
TheCharlatan, line 523: What's up with this semicolon?

Lederstrumpf:
> _ No description provided. _

> Not a bug, because the first if case returns in any case, making the new behavior the same as the old behavior, but still better and more readable with this patch.

general comment @zkao applicable here & elsewhere: it would be helpful if you could describe what you're fixing in a PR, or provide more detail on the features it's adding. I've often detected incorrect assumptions etc. in my thinking when I forced myself to spell out the logic. Incidentally, it's also more respectful of reviewers' time.

TheCharlatan:
This is outdated, if it still is an issue, please follow up with another PR.

PR #641, opened 2022-08-09
Petrinet

zkao:


TheCharlatan:
This is outdated now with the new state machine. If the swap state is supposed to be refactored as a series of implementable traits instead of enums, this PR needs to be re-done from scratch, so I'm closing this now.

PR #642, opened 2022-08-10
doc: move documentation to wiki

h4sh3d:
Documentation is moved to the project's wiki to ease of maintenance.


h4sh3d:
If a dev clone that repo he has an access to the wiki, if you want we can add the wiki git as an indication in the readme so he can also clone the doc.

Lederstrumpf:
> I'm not convinced about this one, because then the docs aren't in a git repo, that a dev clones and can read in his editor. Additionally, it seems fragile to further depend on third party infrastructure (now to read the docs)

It's still markdown in a git repo, just more accessible, in particular for non-devs. No fragility induced by moving it to its own repo where GitHub can present it with a convenient navigation sidepanel.

Lederstrumpf (approved):
ACK - good call to merge our info resources.

On a related (to reviewing this) & fun note, I can really recommend `delta` as a diff engine: https://github.com/dandavison/delta

PR #646, opened 2022-08-11
wallet: improve wallet permission

zkao:


src/walletd/runtime.rs
```diff
@@ -360,8 +360,18 @@ impl Runtime {
                 error!("Msg and source don't have same swap_id, ignoring...");
                 return Ok(());
             }
+
+            Request::Protocol(msg)
```
TheCharlatan, line 364: Why are you adding these permission checks? I don't think we should firewall the services that are on a common service bus from each other.

TheCharlatan: > I disagree, there is no reason to leave an attack surface open for exploiting all the bugs we have left

This "permission" check does not add any actual security in my eyes. Our internal messages are unauthenticated and can be read out, altered and re-routed as an attacker pleases. We already have a firewall in place in peerd that only allows whitelisted messages into the internal message busses and only passes protocol messages to hardcoded services. Any additional internal checks are easily evaded.

That said, if this increases the code robustness, that is a fine reason for me.

src/walletd/runtime.rs
```diff
@@ -360,8 +360,18 @@ impl Runtime {
...
+                if req_swap_id.is_some() && Some(msg.swap_id()) == req_swap_id => {}
             // TODO enter farcasterd messages allowed
-            _ => {}
+            _ if source == ServiceId::Farcasterd => {}
```
TheCharlatan, line 367: This overrides the first match arm. Either remove the specific selection of the TakerCommit protocol message or this arm here.

src/walletd/runtime.rs
```diff
@@ -360,8 +360,18 @@ impl Runtime {
...
+            _ => {
+                error!(
```
TheCharlatan, line 369: Can you move the two error arms so that they follow one another? That would make it easier to parse.

TheCharlatan (changes requested):

TheCharlatan (approved):

TheCharlatan:
Thank you for the fixes :)

PR #647, opened 2022-08-11
request: GetInfo display / discriminate Reveal params and Reveal proof display

zkao:


TheCharlatan:
What is the reason for this change?

h4sh3d:
Correct patch can also be
```diff
-     #[display(Debug)]
```
That prints `GetInfo` with `{}` and `{:?}`.

Why are we redefining display on all enums? I caught some copy/paste errors last time I worked on `swap-cli progress --follow`, maybe we should get rid of all `xxx()` versions and just keep the one we really want to redefine.

PR #655, opened 2022-08-15
Deps: Bump to core 0.5.1

TheCharlatan:


h4sh3d:
To make tests works see: https://github.com/farcaster-project/farcaster-core/pull/293

We have to create a `nightly` feature in node that enables `farcaster_core/nightly` when we compile node with Rust nigthly, i.e. in tests.
```yaml
[features]
nightly = ["farcaster_core/nightly"]
```
Unit tests should then pass because of `--all-features` but we should fix integration tests (`test.yml`) and other files
```
cargo test ${{ matrix.test }} --workspace --all-targets --features all,integration_test --color always -- --ignored --test-threads=1
```
**So:**
1. when you done I can push a patch
2. I patch the CI so we remove code coverage and we get rid of nightly usage

TheCharlatan:
I am for 2. , the code coverage tool is not practical for our code base and I don't think I ever used it.

h4sh3d:
> I am for 2. , the code coverage tool is not practical for our code base and I don't think I ever used it.

Done in https://github.com/farcaster-project/farcaster-node/pull/656

src/rpc/request.rs
```diff
@@ -1023,18 +1021,34 @@ pub struct OfferInfo {
...
-#[derive(Clone, PartialEq, Eq, Debug, Display, StrictEncode, StrictDecode)]
+#[derive(Clone, PartialEq, Eq, Debug, Display)]
 #[cfg_attr(
     feature = "serde",
     derive(Serialize, Deserialize),
     serde(crate = "serde_crate")
 )]
 #[display(TookOffer::to_yaml_string)]
 pub struct TookOffer {
-    pub offerid: PublicOfferId,
```
h4sh3d, line 1034: Maybe, as mentioned in https://github.com/farcaster-project/farcaster-core/pull/292, we can use `[u8; 16]` to avoid manual impl. of encoding. (But with manual From/Into calls in code when we want `Uuid` types)

h4sh3d (commented):
All good, just one comment about type used in struct that requires strict encoding.

src/rpc/request.rs
TheCharlatan: TookOffer is printed as a yaml string, so I want to preserve the Uuid type. I am not sure how a `impl From` would help here.

h4sh3d (approved):
ACK

PR #659, opened 2022-08-16
Walletd: Remove funding information

TheCharlatan:
The funding is handled by swapd exclusively.

Additionally, with the implemented fee estimation, is displays an incorrect amount.


Lederstrumpf (approved):
ACK

PR #662, opened 2022-08-17
Swap: Remove log used while debugging

TheCharlatan:


TheCharlatan:
Tests are not actually failing, it's just the Rust cache not being uploaded.

PR #666, opened 2022-08-18
Swap event processor: Msg and Clt buses (excluding syncer)

zkao:
Messages received by the service buses in `runtime.rs` are now processed in `process_msgs.rs`

`Event`s are created (`event.rs`). They contain the machinery for replying, forwarding, sending msgs etc, in a more abstracted manner with a nice API (barelly used so far, not to modify behavior as of yet)

There are predicates in `swap_state.rs` that inputs Events + State + SyncerState  + Pending requests etc, and compute the variant of the Awaiting enum that encodes abstractly a valid state where some response from the daemon can be elicited. Thus each variant represents an overall state in which that event can elicit a response.

This PR completes all the message processing in the Msg bus and particially for the Ctl bus (Syncer messages missing)

The code moved from `runtime.rs` to `process_msgs.rs` was largely unchanged except when needed. 

Because of the large code move, I signed the commits


src/event.rs
```diff
@@ -0,0 +1,154 @@
...
+    ) -> Result<(), esb::Error<ServiceId>> {
+        self.endpoints.send_to(
+            ServiceBus::Ctl,
+            self.service.clone(),
+            service,
+            message.into(),
+        )
+    }
+
+    pub fn forward(
```
TheCharlatan, line 125: docstring?

src/service.rs
```diff
@@ -326,6 +326,9 @@ where
         }
         Ok(())
     }
+    fn enquirer(&self) -> Option<ServiceId> {
```
TheCharlatan, line 329: What's the idea behind this?

src/swapd/runtime.rs
```diff
@@ -469,8 +491,8 @@ impl esb::Handler<ServiceBus> for Runtime {
         request: Request,
     ) -> Result<(), Self::Error> {
         match bus {
-            ServiceBus::Msg => self.handle_rpc_msg(endpoints, source, request),
-            ServiceBus::Ctl => self.handle_rpc_ctl(endpoints, source, request),
+            ServiceBus::Msg => self.handle_p2p(endpoints, source, request),
```
TheCharlatan, line 494: Can you document somewhere that the Msg bus is used for the p2p comms?

src/swapd/swap_state.rs
```diff
@@ -114,10 +124,287 @@ pub enum SwapCheckpointType {
     CheckpointAlicePreBuy,
 }
 
+pub enum Awaiting {
```
TheCharlatan, line 127: I am a bit confused by the name `Awaiting`, can you comment on that, or describe here what the purpose of this enum is.

src/swapd/swap_state.rs
```diff
@@ -114,10 +124,287 @@ pub enum SwapCheckpointType {
...
+    BobCtlBuySig,
+    BobCtlTxLock,
+    AliceBobCtlTx,
+    AliceBobCtlSweepMonero,
+    AliceBobCtlTerminate,
+    AliceBobCtlSweepBitcoin,
+    AliceBobCtlRpcAbortSimple,
+    BobCtlRpcAbortBob,
+    AliceBobCtlRpcAbortBlocked,
+    AliceBobRpcGetInfo,
```
TheCharlatan, line 152: The `AliceBob` naming is a bit confusing to me. It took a second for me to understand that these are just the ones that are shared between them. Is it possible to just use the shared predicates without the `AliceBob` prefix and similarly for the state evaluation and transition functions?

src/swapd/process_msgs.rs
```diff
@@ -0,0 +1,1040 @@
...
+            // We pass ESB errors forward such that they can fail the swap.
+            // In the future they can be caught here and used to re-iterate sending of the same
+            // message later without swap halting.
+            Err(err @ Error::Esb(_)) => {
+                error!(
+                    "{} due to ESB failure: {}",
+                    "Failing swap".err(),
+                    err.err_details()
+                );
+                self.report_failure_to(
```
TheCharlatan, line 76: If this (and the one below) is really failing the swap and exiting the runtime, it should report this to farcasterd so farcasterd can do the cleanup. 

If you don't want to fix this here and now, it might even be better to just log and not return an error or exit the runtime in order to not change the failure behavior.

src/swapd/runtime.rs
```diff
@@ -873,226 +642,26 @@ impl Runtime {
...
                 | ServiceId::Database
-            ) => {}
-            (Request::AbortSwap, ServiceId::Client(_)) => {}
-            (Request::GetInfo, ServiceId::Client(_)) => {}
+                | ServiceId::Swap(_) // for recursing
+            ) if white_list_ctl_msgs(r) => {
+                self.process(endpoints, source.clone(), request.clone())?;
+                return Ok(())
+            }
+            // FIXME rm clone after syncer correctly handled through self.process
```
TheCharlatan, line 654: Nice - I understand the roadmap here.

TheCharlatan (changes requested):
This gets us very close to swapd as a readable state machine. I just have some preliminary comments for now before I give this a more in depth read.

src/service.rs
TheCharlatan: Would you give your ok to removing it then?

TheCharlatan:
> @TheCharlatan can you please let us know what still needs to be changed on this PR? And how we can proceed in getting it on main?

I'll look through it again now.

src/swapd/process_msgs.rs
```diff
@@ -0,0 +1,1043 @@
...
+    runtime::{
+        aggregate_xmr_spend_view, remote_params_candidate, PendingRequest, PendingRequests,
+        PendingRequestsT,
+    },
+    swap_state::Awaiting,
+    syncer_client::log_tx_received,
+    State,
+};
+
+impl Runtime {
```
TheCharlatan, line 43: The `process_msgs.rs` file only excutes on the runtime. There are now a total of 5 `impl Runtime` within swapd and I don't think this separation makes sense. Either you separate them with abstractions and actually process them with `next()` on the state machine ( StateMachine -> ( process() ) -> StateMachine ) or add them back in the service runtime.

src/swapd/process_msgs.rs
```diff
@@ -0,0 +1,1043 @@
...
+            // Others
+            Awaiting::Unknown => Err(Error::Farcaster(s!("Invalid state for this event"))),
+        }
+    }
+}
+
+/// Ctl bus processing fns
+impl Runtime {
+    fn taker_ctl_take_swap(&mut self, event: Event<Request>) -> Result<(), Error> {
+        if let Request::TakeSwap(InitSwap {
```
TheCharlatan, line 164: I am not happy with this matching. The event matching should happen one layer further up. `taker_ctl_take_swap` can only process `Request::TakeSwap`, so make that the argument. This also applies to all the functions following this one. Their interfaces basically invalidate the work you already did at their call site.

TheCharlatan (changes requested):

TheCharlatan:
The swap state has been refactored into a state machine independent of this PR, so closing this.

PR #673, opened 2022-08-21
report_to() fn in CtlServer trait and make use of it for reporting

zkao:
opening this not to lose the commit

TODO: 
- remove dest from report fns
- report_to should return Vec<ServiceId>


TheCharlatan:
Succeeded by #848

PR #675, opened 2022-08-22
chore: display for rpc msg

h4sh3d:
Based on @zkao works in #647

Not done yet, have to test and make sure the logs are at least as good as before.


h4sh3d:
> just a note that we want readable the logs

Yes. That's why I make sure I don't print any big information in those display impl. Only the data we need to identify what's going on, as you did with reveal: alice/bob/proof, but not the actual content.

PR #682, opened 2022-08-23
Remove configure_me from project

LeoNero:
As far as I saw, `configure_me` is not used anywhere


TheCharlatan (approved):
Indeed its usage was stale. If we ever want to consolidate our configuration again, we should re-add it.

PR #685, opened 2022-08-23
Fix /compose and Dockerfile

LeoNero:
Changes:

- removed `.github/workflows/create-compose-release.yml`
- remove `/compose/Dockerfile`
- fix the `environment` key in `/compose/docker-compose.yml`
- moved `docker-compose.yml` to `/compose` folder
- fix `/Dockerfile`'s CMD by removing the `-x` and `--data-dir` arguments; the `-x` argument was giving the error `error: Invalid value "lnpz://0.0.0.0:9981/?api=esb" for '--ctl-socket <CTL_SOCKET>': invalid server address string 'lnpz://0.0.0.0:9981/?api=esb'`, and the `--data-dir` argument was giving `thread 'main' panicked at 'Unable to access data directory: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }'src/opts.rs:133:42`
- `/compose/docker-compose.yml` uses the latest `farcasterd` tag and creates a volume from `/compose` to `/var/lib/farcaster`, so that the `compose/farcasterd.toml` can be used

Please test it, too! :)


h4sh3d:
`-x` and port forwarding was used to expose the bus outside of the container so a "host" cli could connect to port 9981. But I prefer documenting usage of `docker-compose exec farcaster swap-cli ...` instead of connecting over the network. So we can remove it. And it is not working anymore because of upstream dependency changes.

I'll do more tests before approve and merge, but concept ACK.

Also we we'll have to check to update documentation in wiki.

LeoNero:
@h4sh3d cool. I tried to fix the `-x` but no success... so I decided to remove it and then open for discussion, but you already clarified it. Thanks! I can update the wiki once this is merged

h4sh3d (approved):
I was working on the doc anyway so I did the update.

Now the docker stack is better, using `:latest` is good but we have to complete improvements on the release process and create releases more often so latest doesn't point to a month old node.

E.g. locally I switched `latest` to `main` so I get the updates.

PR #687, opened 2022-08-23
Swap event processor syncer

zkao:


TheCharlatan:
This PR is also outdated with the implementation of the new state machine, so closing it now.

PR #692, opened 2022-08-24
refactor: shell completion generation

h4sh3d:
This PR
- remove the `shell/` folder
- add a `swap-cli completion <shell>` command to generate `swap-cli` auto-completion script for the given shell
- install completion in farcasterd docker image

Build script become very small and build dependencies are reduced to only tonic build (proto compilation).

This is a followup to @LeoNero PRs cleaning up Docker and some unused functionalities.

Difference with main
- in main we generate completion for each bin, now we generate on-demand only for `swap-cli` (but we could add for other too, if just duplicate a bit of code the way it is now)


src/cli/opts.rs
```diff
@@ -249,6 +250,23 @@ pub enum Command {
         /// The destination address receiving the coins.
         destination_address: XmrAddress,
     },
+
+    /// Output shell completion code for the specified shell (bash, zsh or fish)
```
LeoNero, line 254: I think `(bash, zsh, fish, etc)` would describe the options better. Or listening all shells in `src/cli/command.rs`

`This can be done by sourcing it from the .bash_profile.` only applies for bash I think. For example, in fish, there is no need to source anything, only adding a `<binary-name>.fish` file containing the completions to  `~/.config/fish/completions/` is sufficient.

LeoNero:
Cool, this is much cleaner! Only thing I think is missing is the completion for the other binaries.

h4sh3d:
> Cool, this is much cleaner! Only thing I think is missing is the completion for the other binaries.

The only other binary where it might be useful is `farcasterd`. The others are not launched by the user so I think we can omit them.

src/cli/opts.rs
h4sh3d: Agree, but at the same time I wasn't sure how much detailed this should be. Do you want to propose an alternative text?

LeoNero:
@h4sh3d I agree

TheCharlatan (approved):
Very sane change. I never used these completion and generating them after the fact makes sense.

PR #693, opened 2022-08-26
Farcasterd state machines

TheCharlatan:
This commit re-writes the runtime of the farcaster daemon as a collection of deterministic finite automata. It defines traits and types for a state machine that consumes events and manipulates a runtime. The actual automata are documented and described in their respective implementations.
    
Requests pertaining to offers and swaps are processed in the TradeStateMachine. It tracks the lifetime of a single offer and swap starting at either a make offer, take offer or a swap restore. All running trade state machines are held in a collection in the runtime.
    
Similarly, syncer related requests are processed by the SyncerStateMachine and held in a separate collection in the runtime.
    
A function pre-processes the requests by selecting the correct state machine and creating an event to process the request.
    
A number of bugs were discovered and fixed in the process of this re-write. The syncer launching and tracking was still not properly done for swap restore and sweep requests. Some changes in the state were not executed in the right order or after the right request. This led to incorrect reporting of open offers, available services and running swaps, and incorrect handling of existing connections.


src/farcasterd/runtime.rs
```diff
@@ -94,84 +89,52 @@ pub fn run(
...
-    progress_subscriptions: HashMap<ServiceId, HashSet<ServiceId>>,
-    funding_btc: HashMap<SwapId, (bitcoin::Address, bitcoin::Amount, bool)>,
-    funding_xmr: HashMap<SwapId, (monero::Address, monero::Amount, bool)>,
-    checkpointed_pub_offers: List<CheckpointEntry>,
-    restoring_swap_id: HashSet<SwapId>,
-    stats: Stats,
-    config: Config,
-    syncer_task_counter: u32,
-    syncer_tasks: HashMap<TaskId, ServiceId>,
+    pub identity: ServiceId,
```
TheCharlatan, line 115: Add docs describing what each of the fields do.

src/farcasterd/runtime.rs
```diff
@@ -94,84 +89,52 @@ pub fn run(
...
+    pub registered_services: HashSet<ServiceId>,
+    pub public_offers: HashSet<PublicOffer>,
+    pub progress: HashMap<ServiceId, VecDeque<Request>>,
+    pub progress_subscriptions: HashMap<ServiceId, HashSet<ServiceId>>,
+    pub checkpointed_pub_offers: List<CheckpointEntry>,
+    pub stats: Stats,
+    pub config: Config,
+    pub syncer_task_counter: u32,
+    pub trade_state_machines: Vec<TradeStateMachine>,
+    pub syncer_state_machines: Vec<SyncerStateMachine>,
```
TheCharlatan, line 131: Maybe index by task ID?

src/farcasterd/runtime.rs
TheCharlatan: Done aae97400973cf161ea18a46018481f7204518f5b

src/farcasterd/runtime.rs
TheCharlatan: Done aae97400973cf161ea18a46018481f7204518f5b

src/event.rs
```diff
@@ -0,0 +1,97 @@
...
+{
+    /// Move state machine to a next step in response to the provided event.
+    /// At the completion of the cycle the state machine is consumed and `Ok(None)` is returned.
+    fn next(self, event: Event, runtime: &mut Runtime) -> Result<Option<Self>, Error>
+    where
+        Self: Sized;
+}
+
+/// Event changing state machine state, consisting of a certain P2P or RPC `request` sent from some
+/// serivce `source` to the current `service`.
```
Lederstrumpf, line 22: service*

src/farcasterd/syncer_state_machine.rs
```diff
@@ -0,0 +1,327 @@
...
+    syncer: ServiceId,
+    syncer_task: Task,
+    syncer_task_id: TaskId,
+}
+
+pub struct AwaitingSyncerRequest {
+    syncer_task_id: TaskId,
+    source: ServiceId,
+    syncer: ServiceId,
+}
```
Lederstrumpf, line 62: just for readability, would be nice to have the `AwaitingSyncer`/`AwaitingSyncerRequest` fields sorted consistently.

src/farcasterd/trade_state_machine.rs
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+impl StateMachine<Runtime, Error> for TradeStateMachine {
+    fn next(self, event: Event, runtime: &mut Runtime) -> Result<Option<Self>, Error> {
+        match self {
+            TradeStateMachine::StartTaker => transition_to_take_offer(event, runtime),
+            TradeStateMachine::StartMaker => transition_to_make_offer(event, runtime),
+            TradeStateMachine::StartRestore => transition_to_restoring_swapd(event, runtime),
+            TradeStateMachine::MakeOffer(make_offer) => {
+                attempt_transition_to_taker_commited(event, runtime, make_offer)
```
Lederstrumpf, line 161: s/commited/committed/ (applicable to `attempt_transition_to_taker_commited` too)

src/farcasterd/runtime.rs
```diff
@@ -97,84 +83,52 @@ pub fn run(
...
+    pub stats: Stats,                                   // Some stats about offers and swaps
+    pub config: Config, // Configuration for syncers, auto-funding, and grpc
+    pub syncer_task_counter: u32, // A strictly incrementing counter of issued syncer tasks
+    pub trade_state_machines: Vec<TradeStateMachine>, // New trade state machines are inserted on creation and destroyed upon state machine end transitions
+    pub syncer_state_machines: HashMap<TaskId, SyncerStateMachine>, // New syncer state machines are inserted by their syncer task id when sending a syncer request and destroyed upon matching syncer request receival
 }
 
 impl CtlServer for Runtime {}
 
 #[derive(Default)]
```
Lederstrumpf, line 130: ahh! so much :cut_of_meat:  trimmed :100:

once you're done with other changes, reminder to reduce visibility of the fields here, would be nice to have at most `pub(super)` since afaict these are only accessed by `farcasterd::{syncer, trade}_state_machine` (ditto for other funcs in `runtime`)

src/farcasterd/syncer_state_machine.rs
```diff
@@ -0,0 +1,327 @@
...
+        syncer,
+    } = awaiting_syncer_request;
+    match (event.request.clone(), event.source.clone()) {
+        (Request::SyncerEvent(SyncerEvent::SweepSuccess(success)), syncer_id)
+            if syncer == syncer_id && success.id == syncer_task_id =>
+        {
+            if let Some(Some(txid)) = success
+                .txids
+                .clone()
+                .pop()
```
Lederstrumpf, line 270: don't quite understand why this is matching on the last tx only, or receiving a vec of txs at all? don't see why it's not a single tx.

src/farcasterd/runtime.rs
```diff
@@ -943,12 +455,12 @@ impl Runtime {
                     }
                     OfferStatusSelector::InProgress => {
                         let pub_offers = self
```
Lederstrumpf, line 456: incidentally, also more efficient for a long-running node with large histoy of swaps

src/farcasterd/runtime.rs
```diff
@@ -1337,8 +509,8 @@ impl Runtime {
             }
 
             // Returns a unique response that contains the complete progress queue
-            Request::ReadProgress(swapid) => {
```
Lederstrumpf, line 1340: since cleaning up redundant state anyway, this `checkpointed_pub_offers` is quite unwieldy. Seems it's only needed for the trade state machine to get the public offer & trade role when fed a swap id.

Lightly suggest moving the public offer & trade role into the event msg passed to `transition_to_restoring_swapd` as well to get rid of this.

src/farcasterd/runtime.rs
```diff
@@ -1448,202 +615,18 @@ impl Runtime {
...
+                    .iter()
+                    .filter_map(|tsm| tsm.needs_funding_monero())
+                    .collect();
+                let len = funding_infos.len();
+                let res = funding_infos
                     .iter()
-                    .filter(|(_, (_, _, autofund))| !*autofund)
                     .enumerate()
-                    .map(|(i, (swap_id, (address, amount, _)))| {
-                        let mut res = format!(
```
Lederstrumpf, line 1639: very gratifying to see the pattern of replacing external state aggregation by state machine state pattern/predicate matching repeating everywhere :)

src/farcasterd/runtime.rs
```diff
@@ -2080,28 +1149,7 @@ fn launch_swapd(
     )?;
     let msg = format!("New instance of swapd launched with PID {}", child.id());
     debug!("{}", msg);
```
Lederstrumpf, line 1168: great that these are now minimized (of course, still pending a potential merge of swap id & offer uuids)

Lederstrumpf (changes requested):
a few comments, typo pointers, & small requests for now. some unreviewed ground especially in the syncer state machine internals.

I'll leave my bags of praise for later, since this PR's been whispering [this](https://impossiblehq.com/wp-content/uploads/2013/04/Final-Form.jpg) in my ear all evening.

src/farcasterd/runtime.rs
```diff
@@ -1014,315 +489,11 @@ impl Runtime {
             Request::CheckpointList(checkpointed_pub_offers) => {
                 self.checkpointed_pub_offers = checkpointed_pub_offers.clone();
```
Lederstrumpf, line 490: since cleaning up redundant state anyway, this `checkpointed_pub_offers` is quite unwieldy. Seems it's only needed for the trade state machine to get the public offer & trade role when fed a swap id.

Lightly suggest moving the public offer & trade role into the event msg passed to `transition_to_restoring_swapd` as well to get rid of this.

src/farcasterd/runtime.rs
```diff
@@ -1448,202 +614,18 @@ impl Runtime {
...
-
             Request::NeedsFunding(Blockchain::Monero) => {
-                let len = self.funding_xmr.len();
-                let res = self
-                    .funding_xmr
+                let funding_infos: Vec<MoneroFundingInfo> = self
+                    .trade_state_machines
+                    .iter()
+                    .filter_map(|tsm| tsm.needs_funding_monero())
+                    .collect();
```
Lederstrumpf, line 622: very gratifying to see the pattern of replacing external state aggregation by state machine state pattern/predicate matching repeating everywhere :)

src/farcasterd/runtime.rs
```diff
@@ -2042,32 +1140,20 @@ fn syncers_up(
...
-    peerd: ServiceId,
-    report_to: Option<ServiceId>,
+pub fn launch_swapd(
     local_trade_role: TradeRole,
     public_offer: PublicOffer,
-    local_params: Params,
     swap_id: SwapId,
-    remote_commit: Option<Commit>,
-    funding_address: Option<bitcoin::Address>,
 ) -> Result<String, Error> {
```
Lederstrumpf, line 1157: great that these are now minimized (of course, still pending a potential merge of swap id & offer uuids)

src/farcasterd/runtime.rs
```diff
@@ -918,17 +424,22 @@ impl Runtime {
...
             Request::ListOffers(offer_status_selector) => {
                 match offer_status_selector {
                     OfferStatusSelector::Open => {
-                        let pub_offers = self
-                            .public_offers
+                        let open_offers = self
+                            .trade_state_machines
                             .iter()
-                            .filter(|k| !self.consumed_offers_contains(k))
+                            .filter_map(|tsm| tsm.open_offer())
```
Lederstrumpf, line 442: incidentally, also more efficient for a long-running node with large history of swaps

Lederstrumpf (changes requested):
my externally submitted review wasn't in sync with head of PR. resubmitting these comments since they ended up on the wrong lines. apologies for the mess.

src/farcasterd/syncer_state_machine.rs
TheCharlatan: This goes back to the syncer handling a potential excess of transaction outputs for Monero, since Monero transactions can handle only a limited number of inputs. We use the same type to report Success in the bitcoin syncer, so we can just pop the first element, since we don't handle multiple sweep transactions for Bitcoin.

src/event.rs
TheCharlatan: 6b7e3fd59a02e79a24395ebe47706f30ac111483

src/farcasterd/syncer_state_machine.rs
TheCharlatan: 6b7e3fd59a02e79a24395ebe47706f30ac111483

src/farcasterd/trade_state_machine.rs
TheCharlatan: 6b7e3fd59a02e79a24395ebe47706f30ac111483

src/farcasterd/runtime.rs
TheCharlatan: I prefer not add these qualifiers as part of this PR. We don't use them anywhere else in the code where we similarly export structs and member functions, and their benefit is questionable, since we are not writing a library here. I did make sure though that only those functions and fields are pub that need to be in 04997366cc2419fc04fbbd1f8059079eb9d03c10 and afb0647cf9a8ff4c1eba3052fb735d6f54f1973d .

src/farcasterd/runtime.rs
TheCharlatan: The important bit about the `checkpointed_pub_offers` is not necessarily the offer and trade role, but rather that it allows us to immediately report back to the user if the checkpoint does not exist. To get rid of this cache, we'd have to introduce further messages and interim states to the restore workflow in order to report back to the user in time if the checkpoint does not exist. This is not an easy change to tackle, so I'd rather defer this for a separate PR.

src/farcasterd/syncer_state_machine.rs
Lederstrumpf: ok, so it's again there to keep the implementations symmetric - fair

src/farcasterd/runtime.rs
Lederstrumpf: Thanks for cleaning up.

The `super` qualifiers have more purpose in a library since that protects misuse by external projects - the goal here would be to protect development within the node.
But thinking about it now: since the node's segregated into microservices, the natural visibility scope for most `pub` things (and in particular the ones here) is effectively already `pub(super)`.

src/farcasterd/runtime.rs
Lederstrumpf: sure, we can leave it for later.

> it allows us to immediately report back to the user if the checkpoint does not exist

some issues with the restore atm:
if the user hasn't called `list checkpoints` yet and goes straight for `restore checkpoint`, the trade state machine throws `No checkpoint found with given swap id, aborting restore.`, which is misleading for anyone not familiar with the service internals.

imo that tradeoff in favour of reporting back to the user immediately is worse, since in order for the response to their `restore checkpoint` request to have meaning, they have to `list checkpoints` first anyway, which should confirm to them already whether or not the checkpoint exists.

For a message workflow that allows immediate feedback, I think doing
```
cli -> databased : RestoreCheckpoint
if checkpoint-exists {databased -> farcasterd : RestoreCheckpoint} else {databased -> cli: CheckpointNotFound; exit}
farcasterd -> swapd : launch
farcasterd -> databased : RestoreCheckpoint
databased -> walletd : Checkpoint
databased -> swapd: Checkpoint
```
should work - I'll try this out later.

TheCharlatan: I'm already working on it, but will not push to this branch. Yes, the flow you showed here should work.

src/farcasterd/runtime.rs
```diff
@@ -469,69 +261,8 @@ impl Runtime {
...
-                    endpoints.send_to(
-                        ServiceBus::Msg,
-                        self.identity(),
-                        ServiceId::Wallet,
-                        request,
-                    )?;
-                }
-                return Ok(());
-            }
             _ => {
```
h4sh3d, line 264: I find the resulting `match` arms a bit weird now, the main part is in the `_` catch all and we don't use the source anymore

h4sh3d (commented):
I'll give it another round of reading tomorrow. For now just one minor comment.

src/farcasterd/runtime.rs
```diff
@@ -1907,7 +708,316 @@ impl Runtime {
...
+            (Request::LaunchSwap(LaunchSwap { public_offer, .. }), _) => Ok(self
+                .trade_state_machines
+                .iter()
+                .position(|tsm| {
+                    if let Some(tsm_public_offer) = tsm.consumed_offer() {
+                        tsm_public_offer == public_offer
+                    } else {
+                        false
+                    }
+                })
```
Lederstrumpf, line 895: this will forward the request to a swap matching `public_offer` regardless of the swap's state.
tsm.consumed_offer() would for instance also work for `TradeStateMachine::SwapdRunning(SwapdRunning { public_offer, .. })`.

"Luckily", only `attempt_transition_to_swapd_launched` has a matching handler for `Request::LaunchSwap`, but in general, I think it's open how much of the checking `match_request_to_trade_state_machine` should handle:

In principle, all of the validation could happen here already, but I think we're better off only having it handle the minimum (e.g. ensuring that `swap_id` or `public_offer` match, which should be enough to have a unique & compatible match), and then leaving the remaining checks for the state transition function implementations.

But this necessitates that their `event.request` matchers can't ever have catch-alls that return new state, only logging & `None` or the original state like [here](https://github.com/farcaster-project/farcaster-node/pull/693/files#diff-e5f6d9e096b173ed0a096a10871b452ffe63db2298afffdc1b61652bd5f051dfR497-R510).

Lederstrumpf (approved):
tACK.

This is a truly excellent PR - I reckon the best on this repo so far. Despite its size, it remains focused & self-consistent. In particular the tsm cleans up / obsoletes a lot of logic & state manipulation that was previously held together by hopium, and it fills me with hopium for the remainder ;)

I've made some small adjustments in the commits over [here](https://github.com/TheCharlatan/farcaster-node/compare/farcasterdStateMachine...Lederstrumpf:farcaster-node:farcasterdStateMachine-suggestions), @TheCharlatan you can pick&choose from their content what you deem useful.

src/farcasterd/syncer_state_machine.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+fn transition_to_awaiting_syncer_or_awaiting_syncer_request(
+    event: Event,
+    runtime: &mut Runtime,
+) -> Result<Option<SyncerStateMachine>, Error> {
+    let source = event.source.clone();
+    match event.request.clone() {
+        Request::SweepMoneroAddress(sweep_monero_address) => {
```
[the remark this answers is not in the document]
TheCharlatan: Actually, this could be easily achieved with a unified sweep address request. I'll submit a follow-up PR for this one.

src/farcasterd/runtime.rs
TheCharlatan: The function `match_request_to_trade_state_machine` will disappear once the ServiceBus segregation lands as well. As long as we don't have unified identifiers we'll be stuck with this matching though. The way I see it, as little validation as possible should be applied within this auxiliary matching function. Any fine-grained logic should happen in the state machine handlers. Once the identifiers are unified, we can add a trait to each request that returns an identifier of sorts and allows as to immediately select a state machine without any further unpacking of the request.

src/farcasterd/trade_state_machine.rs
```diff
@@ -0,0 +1,1290 @@
...
+        Ok(Some(tsm))
+    } else {
+        Ok(Some(TradeStateMachine::TakerCommit(TakerCommit {
+            peerd,
+            public_offer,
+        })))
+    }
+}
+
+fn attempt_transition_from_take_offer_to_swapd_launched(
```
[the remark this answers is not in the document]
TheCharlatan: Fully agree, I found a nice middle route I think: 7cc52b2b997d5a072a2199ce5c740415ec71beb6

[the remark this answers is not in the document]
Lederstrumpf: yes! that's much better.

src/farcasterd/trade_state_machine.rs
```diff
@@ -0,0 +1,1295 @@
...
+    funding_info: Option<FundingInfo>,
+    auto_funded: bool,
+}
+
+impl StateMachine<Runtime, Error> for TradeStateMachine {
+    fn next(self, event: Event, runtime: &mut Runtime) -> Result<Option<Self>, Error> {
+        match self {
+            TradeStateMachine::StartTaker => transition_to_take_offer(event, runtime),
+            TradeStateMachine::StartMaker => transition_to_make_offer(event, runtime),
+            TradeStateMachine::StartRestore => transition_to_restoring_swapd(event, runtime),
```
Lederstrumpf, line 158: for consistency:
since they all return `Result<Option<TSM>>` like the other `attempt_transition_to_...`, their return is directly forwarded via `.next()`, and to keep them segregated from the new `transition_to_swapd_launched_tsm`, which actually returns a naked `Result<TSM>`, I'd rename these three  `attempt_transition_to...` as we;;.

TheCharlatan: 428c5132925f4f02ed9a3bf6cda34fd7bffd01bf

Lederstrumpf: thanks! I was **just** about to comment the same for the syncer state machine ^^

src/farcasterd/syncer_state_machine.rs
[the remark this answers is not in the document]
Lederstrumpf: tracking in #714

h4sh3d (approved):
Reviewed it while working on #707 which includes this PR. Big ACK!

PR #695, opened 2022-08-28
electrs failures: reattempt `script_get_history(...)`

### LINKED ISSUE #628: Bitcoin Syncer failures not logged in production build
Bitcoin syncer is failing on me regularly (using [ssl://blockstream.info:700]()).
I've had 100 swaps running overnight where the takers disconnected for well over the cancel threshold, and the maker didn't cancel since it wasn't receiving block height updates from the syncer, but also no errors logged for the production build.
Monero syncer was still working, so would maker would also broadcast buy txs, but these would be ignored silently.
Maybe we're getting filtered by blockstream.info?

Anyway: I'll switch to debug logging to identify what went wrong, but whatever the issue is - we should have users be notified in production builds of such issues.

[maker_bitcoin_syncer_debug.log](https://github.com/farcaster-project/farcaster-node/files/9248846/maker_connection_debug.log)

Lederstrumpf:
Closes #628. The failures reported in #628 stem from the `client.transaction_get(&tx_id)` call successfully returning the `tx`, but said `tx_id` not showing up in the script history retrieved via `script_get_history(&tx.output[0].script_pubkey)` yet, which would panic the syncer. This issue's gotten worse with electrs v0.9.x.

Since this would crash the syncer, an error in one swap would break Bitcoin liveness for all running swaps, which should not be happening at all, in particular since liveness on the Bitcoin chain is safety critical.

A mitigation is to reattempt the call after some time delay - this has worked for me without failure, but I'm still tweaking to find the appropriate delay. Can also handle this the same as `script_get_history` returning an error, which would then only reattempt upon a block height change.

### DIFF
--- a/src/syncerd/bitcoin_syncer.rs
+++ b/src/syncerd/bitcoin_syncer.rs
@@ -229,7 +229,7 @@ impl ElectrumRpc {
                                     break;
                                 }
                                 warn!("{:?} should be found in the history if we successfully queried `transaction_get` - reattempting {:?}", &tx_id, attempts);
-                                std::thread::sleep(Duration::from_secs_f32(0.5 * attempts));
+                                std::thread::sleep(Duration::from_secs_f32(0.5 * attempts as f32));
                             }
                             Err(err) => {
                                 trace!(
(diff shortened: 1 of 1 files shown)


Lederstrumpf:
closed in favour of #708

PR #701, opened 2022-08-30
Tests: improve how processes are killed

LeoNero:
This PR does three things:

- renames `setup_clients` to `launch_farcasterd_instances`
- removes `cleanup_processes` and adds a `struct FarcasterdProcess` with a `impl Drop`; this has the advantage that processes will die even if tests fail, which was not happening before
- improve the logic that cleans up processes; it had some problems that processes spawned by `farcasterd` were not being killed

Run the tests (force them to fail too), and in all cases check if any farcaster-related process is running in your machine.

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -11,7 +11,8 @@ mod utils;
 #[tokio::test]
 #[ignore]
 async fn cli_make_offer() {
-    let (farcasterd_maker, data_dir_maker, farcasterd_taker, _) = setup_clients().await;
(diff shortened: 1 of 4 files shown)


h4sh3d:
In #707 we took another approach to kill processes. I like the implementation of `Drop` to cleanup process, it's very rust way of thinking. Do you want to give it a shot and try to port your work based on the state of `main`?

h4sh3d:
Superseded by #707

PR #704, opened 2022-08-30
Restore swap refactor

TheCharlatan:
Instead of keeping track of the available checkpoints in the farcaster runtime, we use the cli to drive execution by first retrieving the checkpoint information for a certain swap id and then triggering the restore workflow.
    
Because the checkpoint list does not need to be persisted in farcasterd anymore, the ckecpoint list is no longer routed through farcasterd.

Based on #693


h4sh3d (approved):
tACK

Tested with the rest of the batch. This new workflow is better and quicker when the list of check-pointed swaps is big.

PR #705, opened 2022-08-30
State machines: Add Executor trait

TheCharlatan:
Adds an executor trait that is useful for a common operation involving state machines: Creating an event and calling next on the state machine and logging its results.

Based on #693


h4sh3d (approved):
tACK

Good refactor, the 2 functions were basically the same. Tested the PR is equivalent feature wise.

Issue #706, opened 2022-08-30
Prepare buses refactor


h4sh3d:
Here is a picture of which microservice uses which buses to communicate with other services.

![Untitled scene](https://user-images.githubusercontent.com/7792896/187479911-9b74e0b0-8e45-4d36-80a6-8239e89e27b9.png)

- `cli` and `grpcd` use the RPC bus to query other services and receive responses
- `farcasterd` is connected to everything
- `peerd` receive and transfer protocol messages on the MSG bus, receive command from `farcasterd` or other services on the CTL bus and listen on RPC bus to `cli` and `grpcd` queries
- `swapd` does the same as `peerd` plus the SYNC bus to receive `syncerd` events
- `walletd` is not directly connected to RPC bus
- `syncerd` only needs CTL to be commanded and SYNC to send events
- `databased` only needs CTL and RPC

TheCharlatan:
Thank you for sketching this so nicely. I've been re-thinking the bus layout again. In my mind, the cli and grpcd (or clients) should have access to the ctl bus as well. In general they are capable of issuing two different kinds of requests. One that triggers a state transition, and one that is purely informational. 

For example the cli can issue things like `info` and `progress` that just read state and don't manipulate anything, while calls like `take` `restore` etc. trigger a long chain of requests and state changes. I therefore suggest to rename the `Rpc` bus to `Info` bus. and give the cli and grpcd access to the `Ctl` bus as well.

This would also makes handling requests manipulating state easier to follow and read.

h4sh3d:
New version of the diagram, also see #707 

![node-buses](https://user-images.githubusercontent.com/7792896/193844448-94d36989-0e5d-4a24-be8c-e8a41000e2d6.png)

PR #707, opened 2022-08-30
Refactor: introduce sync and rpc buses

### LINKED ISSUE #722: Farcasterd panic in `swap_bob_funds_incorrect_amount` test
Test is passing but `farcasterd` is crashing on message

```
thread 'main' panicked at 'attempt to subtract with overflow', src/farcasterd/runtime.rs:181:9
```

which correspond to line
```rust
    pub fn incr_funding_bitcoin_canceled(&mut self) {
        self.awaiting_funding_btc -= 1;
        self.funding_canceled_btc += 1;
    }

```

h4sh3d:
Implements #706, fix #722

This PR introduce two new buses: `sync` and `rpc` and split requests on these new buses.

- [x] Rename module `rpc` into `bus` to have `bus::msg`, `bus::rpc`, `bus::sync` ...
- [x] Migrate cli over `rpc` bus
- [x] Migrate syncers over `sync` bus
- [x] Only have `Rpc`, `Ctl`, `Msg`, and `Sync` on the primary `Request` enum


h4sh3d:
commit `refactor: start splitting request messages` starts splitting the `Request` enum into different enums for each buses.

`swap-cli info` has been fully migrated over the RPC bus; the other cli calls now fail because the rest is still over the CTL bus so CI will not pass.

h4sh3d:
Next steps for this PR to then go into review/heavy test phase next week:
- [ ] check documentation on encoding and `API` usage for messages
- [ ] chose correct numbers for each message (if `API` derive is maintained above)
- [ ] maybe add new message types to allow strict filtering of types on each buses; `Request::Rpc` only on `Rpc` bus, `Ctl` on `Ctl`, etc.

src/bus/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+// the public domain worldwide. This software is distributed without
+// any warranty.
+//
+// You should have received a copy of the MIT License
+// along with this software.
+// If not, see <https://opensource.org/licenses/MIT>.
+
+pub mod ctl;
+pub mod msg;
+pub mod rpc;
```
Lederstrumpf, line 17: what's your take on renaming this bus to `Info`, as suggested by @TheCharlatan before, either in this PR or in followup?
https://github.com/farcaster-project/farcaster-node/issues/706#issuecomment-1231899553

I think that would make sense, given that you've segregated the cli instructions onto the `Ctl` bus already.

h4sh3d: I'd prefer doing it in another PR at this stage. I kept `Rpc` because I wasn't convinced about `Info` as I was at the beginning, but `Rpc` is also misleading and I agree we should find another name for it.

src/syncerd/runtime.rs
```diff
@@ -104,12 +107,18 @@ impl esb::Handler<ServiceBus> for Runtime {
...
-            ServiceBus::Msg => self.handle_rpc_msg(endpoints, source, request),
-            ServiceBus::Ctl => self.handle_rpc_ctl(endpoints, source, request),
-            ServiceBus::Bridge => self.handle_bridge(endpoints, source, request),
+        match (bus, request) {
+            (ServiceBus::Msg, request) => self.handle_msg(endpoints, source, request),
+            // Control bus for internal command
+            (ServiceBus::Ctl, BusMsg::Ctl(req)) => self.handle_ctl(endpoints, source, req),
+            // User issued command RPC bus, only accept BusMsg::Rpc
+            (ServiceBus::Rpc, BusMsg::Rpc(req)) => self.handle_rpc(endpoints, source, req),
+            // Syncer event bus
```
Lederstrumpf, line 118: comment here that this is not consistent with the diagram in
https://github.com/farcaster-project/farcaster-node/issues/706#issue-1355968942.
I'm totally ok with syncerd talking sending info via RPC bus - just making sure
that this is intentional. Would be nice if we have a diagram that's consistent
with the changes in this PR in the docs (likewise accounting for the fact that
cli sends instructions via CTL bus).

h4sh3d: Yes the diagram must be updated, and the refactor also contain some bus handling (that are empty) that will be removed while simplifying the Hello msg at the start.

h4sh3d: I'll post an update on this PR before moving it to the diagram folder in the source code.

src/bus/msg.rs
```diff
@@ -0,0 +1,135 @@
...
+    swap::SwapId,
+};
+use internet2::Api;
+use strict_encoding::{StrictDecode, StrictEncode};
+
+#[derive(Clone, Debug, Display, Api, StrictDecode, StrictEncode)]
+#[api(encoding = "strict")]
+#[display(inner)]
+#[non_exhaustive]
+pub enum Msg {
```
Lederstrumpf, line 17: given that these are all **protocol** messages, I'd find it ideal if we could find a name that captures this - this name is generic and can be confused with something like `Info` too.
Maybe `Protocol`/`Ptl`/`Prot`/`PtclMsg`/`PrtclMsg`, or fully verbose `ProtocolMsg`.
What's your take on this @h4sh3d?

Lederstrumpf (commented):
largely done with review, in particular on high level. Submitting these comments already now so you can think about them while I transition to the lower bang-for-your-buck nitpicking phase of reviewing ;)

With all of my comments so far, what I should stress is that they can all be followups to this PR (if you agree with the suggestions) - farcaster's performing extremely well on this PR's state (+ electrum_client upgrade), so I'm eager to get this in to unblock merge of other PRs.

src/bus/mod.rs
Lederstrumpf: > I'd prefer doing it in another PR at this stage.

:+1:

> I kept Rpc because I wasn't convinced about Info as I was at the beginning, but Rpc is also misleading and I agree we should find another name for it.

To be clear here: you're not fully satisfied with `Info` as a name either and would rather find another name? My comment https://github.com/farcaster-project/farcaster-node/pull/707#discussion_r986829266 is related to this discussion.

src/bus/ctl.rs
```diff
@@ -0,0 +1,486 @@
...
+use crate::bus::rpc::{AddressSecretKey, OfferStatus};
+use crate::bus::BusMsg;
+use crate::swapd::CheckpointSwapd;
+use crate::syncerd::SweepAddressAddendum;
+use crate::walletd::runtime::CheckpointWallet;
+use crate::{Error, ServiceId};
+
+#[derive(Clone, Debug, Display, From, NetworkEncode, NetworkDecode)]
+#[non_exhaustive]
+pub enum Ctl {
```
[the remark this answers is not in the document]
h4sh3d: The goal is to have `2.` with `Ctl` be the execution-relevant bus, your example with `Success`, etc. is half correct, this part needs a better refactor I left outside of this PR but:
- services, like `swapd`, needs to update the progress of a swap in `farcasterd`, so they use `Ctl` to instruct changes in fcd
- clients use the `Rpc` bus to read the progress

This is why we have such types in both buses

src/bus/msg.rs
h4sh3d: Agree that's not optimal. This matches LNP Node... kinda because in LNP it uses Bolt and Bifrost. A simple one could be `P2pMsg`.

Also in the following PR I'd like to rename `Ctl` into `CtlMsg` and have `BusMsg::Ctl(CtlMsg::...)`, etc. We already have it for `SyncMsg` because `Sync` is too close to some primitive names in Rust language.

- https://github.com/LNP-WG/lnp-node/blob/master/src/bus/mod.rs

src/bus/mod.rs
h4sh3d: To be honest I don't know why I don't like `Info` anymore... so we can go ahead with `Info` :see_no_evil: 

It makes sense to have `Info` for "ready only" and `Ctl` for "mutations", and there is no real difference if it is client/service or service/service. But ACL is easier to implement if "mutations" only happens on `Ctl`.

Lederstrumpf: opened #732 to track

h4sh3d:
This PR is a big step toward this architecture:

node-buses

Currently some services are listening to some buses without using it, this needs to be cleanup in the next PR. Also, `walletd` currently make no usage of the `Info` (i.e. `Rpc`) bus, but we can add it if we want to query some state later.

src/bus/msg.rs
Lederstrumpf: > A simple one could be P2pMsg.

I'd rather break rust conventions here and do `P2PMsg`/`p2pMsg` 🙈 
Maybe `PeerMsg`.
It would actually be nice if these message would their actual destination or origin - that could make handling easier to follow (something like `FromPeerMsg` or `ToPeerMsg`). What's less nice about that is that this would necessitate converting these types inside peerd. 

> Also in the following PR I'd like to rename Ctl into CtlMsg and have BusMsg::Ctl(CtlMsg::...), etc. We already have it for SyncMsg because Sync is too close to some primitive names in Rust language.

great idea!

src/bus/ctl.rs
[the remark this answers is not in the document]
Lederstrumpf: got it, so having those messages on the `Ctl` bus too makes sense for upcoming ACL implementation since they mutate state in a service too, even if it's not relevant for protocol execution.

My current position is to leave as is and to further split up buses by services-only/including-external-agents only if our interfaces grow big enough to hurt code readability, or we want to impose very distinct ACLs for services-only/including-external-agents.

src/farcasterd/runtime.rs
```diff
@@ -106,16 +106,16 @@ pub fn run(
...
+    wallet_token: Token,                         // Set on Runtime instantiation
+    started: SystemTime,                         // Set on Runtime instantiation
     node_secret_key: Option<SecretKey>, // Set by Keys request shortly after Hello from walletd
     node_public_key: Option<PublicKey>, // Set by Keys request shortly after Hello from walletd
     pub listens: HashSet<InetSocketAddr>, // Set by MakeOffer, contains unique socket addresses of the binding peerd listeners.
     pub spawning_services: HashSet<ServiceId>, // Services that have been launched, but have not replied with Hello yet
     pub registered_services: HashSet<ServiceId>, // Services that have announced themselves with Hello
     pub public_offers: HashSet<PublicOffer>, // The set of all known public offers. Includes open, consumed and ended offers includes open, consumed and ended offers
-    progress: HashMap<ServiceId, VecDeque<Request>>, // A mapping from Swap ServiceId to its sent and received progress requests
+    progress: HashMap<ServiceId, VecDeque<ProgressStack>>, // A mapping from Swap ServiceId to its sent and received progress messages (Progress, Success, Failure)
```
Lederstrumpf, line 118: :+1:

src/farcasterd/runtime.rs
```diff
@@ -134,8 +134,8 @@ pub struct Stats {
     punish: u64,
     abort: u64,
     initialized: u64,
-    awaiting_funding_btc: u64,
-    awaiting_funding_xmr: u64,
+    awaiting_funding_btc: HashSet<SwapId>,
+    awaiting_funding_xmr: HashSet<SwapId>,
```
Lederstrumpf, line 138: much less error prone & expressive as information source :)

I agree to keep the others 
https://github.com/farcaster-project/farcaster-node/blob/feeb691c737fe54815c00ef849c8b7b1fb05e664/src/farcasterd/runtime.rs#L139-L142
as `u64` to avoid state bloat for long running node.

src/farcasterd/runtime.rs
```diff
@@ -151,35 +151,44 @@ impl Stats {
...
     }
-    pub fn incr_awaiting_funding(&mut self, blockchain: &Blockchain) {
+
+    pub fn incr_awaiting_funding(&mut self, blockchain: &Blockchain, swapid: SwapId) {
         match blockchain {
-            Blockchain::Monero => self.awaiting_funding_xmr += 1,
-            Blockchain::Bitcoin => self.awaiting_funding_btc += 1,
-        }
+            Blockchain::Monero => self.awaiting_funding_xmr.insert(swapid),
+            Blockchain::Bitcoin => self.awaiting_funding_btc.insert(swapid),
```
Lederstrumpf, line 162: nit: since we've had issues with inconsistent incrementing/decrementing of funding info, I'd log warnings here if `swapid` is already present (likewise for removals).

src/peerd/runtime.rs
```diff
@@ -314,12 +311,18 @@ impl esb::Handler<ServiceBus> for Runtime {
...
-            ServiceBus::Msg => self.handle_rpc_msg(endpoints, source, request),
-            ServiceBus::Ctl => self.handle_rpc_ctl(endpoints, source, request),
-            ServiceBus::Bridge => self.handle_bridge(endpoints, source, request),
+        match (bus, request) {
+            // Peer-to-peer message bus
+            (ServiceBus::Msg, BusMsg::Msg(req)) => self.handle_msg(endpoints, source, req),
+            // Control bus for issuing internal commands
+            (ServiceBus::Ctl, BusMsg::Ctl(req)) => self.handle_ctl(endpoints, source, req),
+            // User issued command RPC bus, only accept BusMsg::Rpc
+            (ServiceBus::Rpc, BusMsg::Rpc(req)) => self.handle_rpc(endpoints, source, req),
```
Lederstrumpf, line 322: these comments don't match how the buses are used anymore

src/peerd/runtime.rs
```diff
@@ -334,62 +337,69 @@ impl esb::Handler<ServiceBus> for Runtime {
...
-                        )?;
-                        warn!("Exiting peerd");
-                        std::process::exit(0);
-                    }
-
-                    while let Err(err) = self.reconnect_peer() {
-                        warn!("error during reconnection attempt: {}", err);
-                    }
-                }
+        // 1. Check permissions
```
Lederstrumpf, line 346: we're not checking permissions here (yet)

src/service.rs
```diff
@@ -91,11 +95,17 @@ impl FromStr for ClientName {
...
 pub struct ServiceConfig {
-    /// ZMQ socket for lightning peer network message bus
+    /// ZMQ socket for peer-to-peer network message bus
     pub msg_endpoint: ServiceAddr,
 
     /// ZMQ socket for internal service control bus
     pub ctl_endpoint: ServiceAddr,
+
+    /// ZMQ socket for remote procedure call bus
+    pub rpc_endpoint: ServiceAddr,
```
Lederstrumpf, line 105: ditto on documentation, but can also be left for `Info` followup

src/swapd/runtime.rs
```diff
@@ -428,12 +435,19 @@ impl esb::Handler<ServiceBus> for Runtime {
...
-            ServiceBus::Msg => self.handle_rpc_msg(endpoints, source, request),
-            ServiceBus::Ctl => self.handle_rpc_ctl(endpoints, source, request),
-            _ => Err(Error::NotSupported(ServiceBus::Bridge, request.get_type())),
+        match (bus, request) {
+            // Peer-to-peer message bus
+            (ServiceBus::Msg, request) => self.handle_msg(endpoints, source, request),
+            // Control bus for internal command
+            (ServiceBus::Ctl, request) => self.handle_ctl(endpoints, source, request),
+            // User issued command RPC bus, only accept BusMsg::Rpc
+            (ServiceBus::Rpc, BusMsg::Rpc(req)) => self.handle_rpc(endpoints, source, req),
```
Lederstrumpf, line 446: nit: see comments comment elsewhere

Lederstrumpf (approved):
great PR! also nice to have a bunch of redundant code removed :)

tACK @https://github.com/Lederstrumpf/farcaster-node/commit/feeb691c737fe54815c00ef849c8b7b1fb05e664, which is built on this.
```
[2022-10-04T15:52:26Z INFO  farcaster_node::farcasterd::runtime] Swapped(290) | Refunded(1) / Punished(0) | Aborted(0) | Initialized(324) / AwaitingFundingXMR(6) / AwaitingFundingBTC(0) / FundedXMR(157) / FundedBTC(153) / FundingCanceledXMR(0) / FundingCanceledBTC(0)
```
Where the only refund is intentional.
The only nits I'd really want included now is to include the diagram https://github.com/farcaster-project/farcaster-node/pull/707#issuecomment-1267079237 in the repo, and https://github.com/farcaster-project/farcaster-node/pull/707#discussion_r986799580. Other open discussions I've opened here should be followups since they're less immediately important than other work on the node.

src/peerd/runtime.rs
h4sh3d: Not a new line, was originally in code refactored. I’ll add it as a todo

src/farcasterd/runtime.rs
h4sh3d: Also thought to switch to sets for all of them, but wanted to discuss it in another PR.

IMO it's not a problem even for long running nodes, and we could use another type (a queue for example) to keep the size constrained.

src/farcasterd/runtime.rs
h4sh3d: Done in [c6ab6df](https://github.com/farcaster-project/farcaster-node/pull/707/commits/c6ab6df1a2044e2a5a4056e677dfffd867683d92)

src/peerd/runtime.rs
h4sh3d: Done in [f28c0b2](https://github.com/farcaster-project/farcaster-node/pull/707/commits/f28c0b26a8a366ff8ad73e6285c37ae985c9a76c)

src/swapd/runtime.rs
h4sh3d: Done in [f28c0b2](https://github.com/farcaster-project/farcaster-node/pull/707/commits/f28c0b26a8a366ff8ad73e6285c37ae985c9a76c)

src/peerd/runtime.rs
h4sh3d: Done in [f87e4de](https://github.com/farcaster-project/farcaster-node/pull/707/commits/f87e4de068e9904b24cbb6fbe5ea4776496ddd2d)

src/bus/rpc.rs
```diff
@@ -0,0 +1,428 @@
...
+    Bitcoin {
+        address: bitcoin::Address,
+        secret_key: bitcoin::secp256k1::SecretKey,
+    },
+    Monero {
+        address: monero::Address,
+        view: monero::PrivateKey,
+        spend: monero::PrivateKey,
+    },
+}
```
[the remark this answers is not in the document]
h4sh3d: Done in [7ca9191](https://github.com/farcaster-project/farcaster-node/pull/707/commits/7ca919117946f6d64d388ce2525da6f5c89e291c)

src/walletd/runtime.rs
```diff
@@ -293,11 +293,11 @@ impl esb::Handler<ServiceBus> for Runtime {
...
             // Peer-to-peer message bus
             ServiceBus::Msg => self.handle_msg(endpoints, source, request),
-            // Control bus for internal command
+            // Control bus for issuing control commands
             ServiceBus::Ctl => self.handle_ctl(endpoints, source, request),
-            // Syncer event bus
+            // Syncer event bus for blockchain tasks and events
             ServiceBus::Sync => self.handle_sync(endpoints, source, request),
-            // All other buses are not supported in walletd
+            // All other pairs are not supported
```
Lederstrumpf, line 300: this handler's only matching on the bus, not BusMsg flavour. Walletd will match `(ServiceBus, BusMsg)` pair ultimately too, right?

h4sh3d: Yes

PR #711, opened 2022-08-31
Farcasterd connect refactor

TheCharlatan:
This is done with the goal of making the processing of make and take offer easier to read by removing most of the and_then statements and unnesting some of the logic.

Coincidentally, this also solves a bug where the peer services are not removed from the spawning_services HashSet and where the connect function might double-spawn a connection that has not said Hello to farcasterd yet.

Lastly, the swap state machine now checks during the transition from SwapdLaunched to SwapdRunning that that the peer connection is up and available before initializing the swap.


h4sh3d (approved):
tACK

Over the last week I've encounter only 1 disconnect, on Monday this week, my client took 3 attempts and reconnected successfully to then complete the swap.

PR #713, opened 2022-08-31
Syncer: normalize treatment between different failure cases when polling transactions

TheCharlatan:
Ensure that once a transaction is found it is treated by minimum as unconfirmed.

### DIFF
--- a/src/syncerd/bitcoin_syncer.rs
+++ b/src/syncerd/bitcoin_syncer.rs
@@ -211,31 +211,25 @@ impl ElectrumRpc {
             match self.client.transaction_get(&tx_id) {
                 Ok(tx) => {
                     debug!("Updated tx: {}", &tx_id);
-                    let history_res;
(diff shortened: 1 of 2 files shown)


src/syncerd/bitcoin_syncer.rs
```diff
@@ -211,31 +211,25 @@ impl ElectrumRpc {
...
-                                    debug!(
-                                        "{} should be found in the history if we successfully queried `transaction_get` for it",
-                                        &tx_id
-                                    );
-                                    continue;
-                                }
-                            }
-                        }
+                                .position(|history_entry| history_entry.tx_hash == tx_id)
+                                .map(|pos| history.remove(pos))
```
Lederstrumpf, line 225: hah! nice trick to get rid of the `history_res` binding.

Lederstrumpf (approved):
tACK

PR #725, opened 2022-09-28
fix sweep failures on intermittent daemon irresponsiveness

Lederstrumpf:
as reported on irc, all sustained maker/taker relationships would eventually halt. this was primarily down to monero sweep sometimes failing: if the daemon would intermittently not respond, the sweep would fail and not be reattempted (this lack of retry was introduced by a code refactor). Thus the swap would never finish for Bob, only Alice.

now sweeps are again always reattempted when this occurs.


h4sh3d (approved):
ACK

PR #731, opened 2022-10-04
Feat: improve swap-cli commands

h4sh3d:
Based on #707 

This PR introduce the implementation of `info <blockchain> <network>` and `list-tasks <blockchain> <network>` to return the list of tasks a _syncer has received_ (and not the current list of tasks, this need a bit more work because of the current syncer state).

It also extend the `info` command to return the `stats` that are printed in the logs (number of funded, swaped, canceled, etc).


h4sh3d:
Rebased on main.

PR #738, opened 2022-10-12
feat: log as info level p2p msg sent and received per swap

h4sh3d:
Log outgoing and incoming p2p protocol message in a friendly manner with the corresponding swap id. Next step is to add them in the progress stack too, but I let it for another PR that improve the overall progress feature.


TheCharlatan (approved):
Code Review and light testing ACK

I like these changes, the displays were kind of insane before.

PR #739, opened 2022-10-15
Feature: Extend Grpc implementation

TheCharlatan:
This pull requests implements additional endpoints for listing peers and checkpoint entries, restoring checkpoints, making and taking offers, revoking offers, aborting swaps, and sweeping coins. 

The grpc server can now handle sending to arbitrary services. Additionally, sending messages back to the client has been extended for including grpc as a possible client, where applicable.

In order to route all possible grpc request messages across all possible service busses through the grpc daemon bridge, a separate BridgeMsg message type was added, acting as a wrapper for the messages and their destination.

Getting full test coverage on this PR is nigh impossible, since there are an abundance of failure paths. This PR does extend the tests though by adding a happy path test for each possible grpc request.

I'd also like to request additional scrutiny of the chosen protobuf types. There are a lot of primitive types that might have better alternatives. I especially dislike the excessive usage of strings. The `needs_funding` and `progress` responses are currently only strings that need to be additionally interpreted and parsed by grpc clients. However, these calls should be refactored outside of this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,9 +141,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "axum"
-version = "0.5.13"
+version = "0.5.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b9496f0c1d1afb7a2af4338bbe1d969cddfead41d87a9fb3aaa6d0bbc7af648"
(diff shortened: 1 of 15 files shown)


TheCharlatan:
I pushed another commit to endpoints for decoding an offer (`offer_info`) and another endpoint for getting the `SwapInfo`. Also includes tests for each.

h4sh3d:
Is it possible to also return the peer we'd connect to in `OfferInfoResponse`, whit that I can match what we have on farcaster.dev

TheCharlatan:
> Is it possible to also return the peer we'd connect to in `OfferInfoResponse`, whit that I can match what we have on farcaster.dev

Done

PR #740, opened 2022-10-17
Fix(syncer_client): remove dangerous unwrap in handle tx confs

h4sh3d:
Fix issue reported by @Lederstrumpf on IRC

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/swapd/syncer_client.rs:368:35
```

This PR change the printing logic of this dangerous `unwrap` (the one before is safe as the condition checks for `is_some`) to not print the variable `confirmations`. Previous code checked if the tx was inserted in the final tx map, now if retrieve the value in the map and print `final` or `non-final`.

A general comments on this piece of code: the map structure is not really used as I only found `insert(txlabel, true)` and nothing else, this is equivalent to simply using a set.

### DIFF
--- a/src/swapd/syncer_client.rs
+++ b/src/swapd/syncer_client.rs
@@ -359,14 +359,16 @@ impl SyncerState {
                     "confirmations".bright_green_bold()
                 );
                 self.tasks.final_txs.insert(*txlabel, true);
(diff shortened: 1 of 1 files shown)


TheCharlatan (approved):
Code Review ACK 844d09054dd09979c4127602ef40efe2b1c2ed6e

PR #744, opened 2022-11-01
Bugfix: Use refund address for sweeping bitcoin through walletd

TheCharlatan:
This allows us to circumvent any dangers incurred from getting the btc addresses.


h4sh3d (approved):
cACK

PR #748, opened 2022-11-03
Syncer: Check for dust amount in sweep bitcoin method

TheCharlatan:
This is a quick fix for the problem reported by @Lederstrumpf on IRC:

I just ran into this: bitcoin syncer panics when sweeping if fee estimate exceeds value of amount to be sweeped:
https://github.com/farcaster-project/farcaster-node/blob/26a46f57643c9dee92d565119fd920364544a480/src/syncerd/bitcoin_syncer.rs#L458

```
thread 'tokio-runtime-worker' panicked at 'attempt to subtract with overflow', src/syncerd/bitcoin_syncer.rs:458:35
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


h4sh3d (approved):
ACK

PR #754, opened 2022-11-05
Cli: Handle routing errors when routing through farcasterd

TheCharlatan:
This commit introduces error handlign in the case where the target services of certain cli calls do not exist. This may happen when aborting a non-running swap, or when querying the info status of a non-existing daemon.


h4sh3d (approved):
ACK

PR #755, opened 2022-11-05
Cli: Implement endpoint for checking syncer health

TheCharlatan:
This adds a cli command "health-check" for checking the configuration and required environment (e.g. the to-be-connected coin daemons, wallets, etc.) of a syncer.

To this end, this commit also adds a HealthCheck event that can be dispatched to a syncer. The various syncers make simple calls to the respective coin daemons and report their success back in the form of a HealthResult event.

The "health-check" command makes use of the syncer state machine in farcasterd to startup syncers and route calls back to the client.

### DIFF
--- a/src/bus/ctl.rs
+++ b/src/bus/ctl.rs
@@ -1,6 +1,7 @@
 use std::fmt::{self, Debug};
 use std::str::FromStr;
 
+use farcaster_core::blockchain::Network;
 use farcaster_core::{
     blockchain::Blockchain,
     role::TradeRole,
@@ -18,7 +19,7 @@ use crate::bus::{
     AddressSecretKey, CheckpointEntry, Failure, OfferStatusPair, OptionDetails, Outcome, Progress,
 };
 use crate::swapd::CheckpointSwapd;
-use crate::syncerd::SweepAddressAddendum;
+use crate::syncerd::{Health, SweepAddressAddendum};
 use crate::walletd::runtime::CheckpointWallet;
 use crate::{Error, ServiceId};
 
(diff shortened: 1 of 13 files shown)


Lederstrumpf (approved):
tACK @ 6dfa15511d5a0a641d51001744e2c6e5b66d9b62. My only improvement suggestion on the UX side is to provide an optional arg to `health-check` to filter by network, such as `health-check mainnet` (or `health-check -n mainnet` to maintain the same structure as `make`). It's only a minor nuisance though to see errors being thrown for networks you're not interested in, so it would be a low priority enhancement at best.

TheCharlatan:
> tACK @ https://github.com/farcaster-project/farcaster-node/commit/6dfa15511d5a0a641d51001744e2c6e5b66d9b62. My only improvement suggestion on the UX side is to provide an optional arg to health-check to filter by network, such as health-check mainnet (or health-check -n mainnet to maintain the same structure as make). It's only a minor nuisance though to see errors being thrown for networks you're not interested in, so it would be a low priority enhancement at best.

I'll do this in a follow-up. It's pretty easy to add this, needs code changes in the cli and grpc server only.

Lederstrumpf:
reference for later before I merge: I've retried this on the parallel takers, and there had a monero-wallet-rpc that was really slow for some reason. The health check stalled for over 90 seconds multiple times then before returning `connection error: Connection reset by peer (os error 104)`. Issue disappeared when I restarted monero-wallet-rpc - shoulda kept it around for reproducibility >:(.

> I'll do this in a follow-up. It's pretty easy to add this, needs code changes in the cli and grpc server only.

thanks - when you do this, can you also reduce await timeouts for the syncers (at least monero)?

PR #759, opened 2022-11-08
Peerd: Use message receipts to confirm receival

TheCharlatan:
Peer messages sent to a remote peer are now cached until the remote peer has replied with a corresponding receipt.

If the connection fails in a listener spawned peerd, the cache is emptied before peerd shuts down and the messages are sent back to their respective origin (swapd). Swapd then processes them into its own cache.

If the connection fails in a connecting peerd, the cache is only emptied once the reconnect has been a success. Messages are not sent back to their origin.

### DIFF
--- a/src/bus/ctl.rs
+++ b/src/bus/ctl.rs
@@ -13,7 +13,7 @@ use bitcoin::Transaction;
 use internet2::addr::{InetSocketAddr, NodeAddr};
 use strict_encoding::{NetworkDecode, NetworkEncode};
 
-use crate::bus::p2p::{Commit, TakerCommit};
+use crate::bus::p2p::{Commit, PeerMsg, TakerCommit};
 use crate::bus::{
     AddressSecretKey, CheckpointEntry, Failure, OfferStatusPair, OptionDetails, Outcome, Progress,
 };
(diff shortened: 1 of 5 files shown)


src/bus/ctl.rs
```diff
@@ -22,6 +22,8 @@ use crate::syncerd::SweepAddressAddendum;
 use crate::walletd::runtime::CheckpointWallet;
 use crate::{Error, ServiceId};
 
+use super::p2p::PeerMsg;
```
h4sh3d, line 25: Move to line 16

src/peerd/runtime.rs
```diff
@@ -327,7 +333,7 @@ impl esb::Handler<ServiceBus> for Runtime {
             // RPC command bus, only accept BusMsg::Info
             (ServiceBus::Info, BusMsg::Info(req)) => self.handle_info(endpoints, source, req),
             // Internal peerd bridge for inner communication, accept all type of request
```
h4sh3d, line 335: Fix comment, not all type of req anymore

h4sh3d (changes requested):
cACK

Only some minor changes

src/peerd/runtime.rs
TheCharlatan: Done

TheCharlatan:
@h4sh3d force pushed the required changes.

TheCharlatan:
I pushed a commit to remove the swapd request whitelisting.

h4sh3d (approved):

PR #764, opened 2022-11-09
Cli: Implement endpoint for listing funding addresses

TheCharlatan:


h4sh3d:
Is there a simple way to add the swap id for each address, if yes this could be done in a follow up PR.

This would allow the GUI to re-conciliate the funding addresses with the swaps and display them together.

TheCharlatan:
> Is there a simple way to add the swap id for each address, if yes this could be done in a follow up PR.

Yes, with https://github.com/farcaster-project/farcaster-node/pull/782 the data is in the database. I can add it in this PR, or in a follow however you prefer.

h4sh3d:
> I can add it in this PR, or in a follow however you prefer.
Perfect! Then I suggest you add it in this one already, it'll be pretty easy to review everything at once

TheCharlatan:
Ok, the swap id is now part of the returned information.

h4sh3d (approved):

PR #765, opened 2022-11-09
Enforce msg and bus coupling

TheCharlatan:
Based on #739


h4sh3d (commented):
We can also fix the `send_ctl` function (or remove if not used), it takes `BusMsg` and should take `CtlMsg`.

Maybe we also want to add `..wallet_{ctl,info,peer}` helpers?

TheCharlatan:
> We can also fix the `send_ctl` function (or remove if not used), it takes `BusMsg` and should take `CtlMsg`.
> 
> Maybe we also want to add `..wallet_{ctl,info,peer}` helpers?
> 
> Otherwise LGTM

I did not want to go overboard here. These changes here were mostly targeted at the state machines. Refactoring all the endpoints usage was intended for another PR.

h4sh3d (approved):

PR #766, opened 2022-11-09
Config: Define global static bind address

TheCharlatan:
Defines a static bind address that will be used for all listens of a running farcaster instance.


src/config.rs
```diff
@@ -28,13 +28,17 @@ pub const FARCASTER_TESTNET_ELECTRUM_SERVER: &str = "ssl://blockstream.info:993"
...
+
 #[derive(Deserialize, Serialize, Debug, Clone)]
 #[serde(crate = "serde_crate")]
 pub struct Config {
     /// Farcasterd configuration
     pub farcasterd: Option<FarcasterdConfig>,
     /// Syncer configuration
     pub syncers: Option<SyncersConfig>,
+    /// Bind address
+    pub bind_addr: String,
```
h4sh3d, line 41: This could go under farcasterd or a new peerd section, what do you think?

TheCharlatan:
> Adding a required field after config got generated will break and force the user to update is config.

That's clear to me, but why does it break if there is no config present?

TheCharlatan:
I'm not too keen on this change anymore. It robs us of a bunch of flexibility. I'll think some more about it and will eventually report back.

h4sh3d:
I propose we take the opportunity here to also change the default port to `7067` which is the decimal repr of "FC". This port number as no entry on iana.org :)

h4sh3d:
I'll rebase and propose a commit for the new default port.

TheCharlatan:
> I'll rebase and propose a commit for the new default port.

Ok, I'll leave my hands off this one.

h4sh3d (approved):
ACK

Waiting on @TheCharlatan ACK and #792

TheCharlatan:
ACK , just left a small additional commit to relax the log level again.

PR #769, opened 2022-11-11
Add richer data to the swap info

TheCharlatan:
Adds the trade role, swap role, and node id of the connected counterparty to the swap Runtime and the SwapInfo structs.

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -84,12 +84,8 @@ pub fn run(
     };
 
     let init_state = match local_swap_role {
-        SwapRole::Alice => State::Alice(AliceState::StartA {
-            local_trade_role: local_trade_role.clone(),
-        }),
-        SwapRole::Bob => State::Bob(BobState::StartB {
-            local_trade_role: local_trade_role.clone(),
-        }),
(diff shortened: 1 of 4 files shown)


src/swapd/runtime.rs
```diff
@@ -81,8 +84,12 @@ pub fn run(
     };
 
     let init_state = match local_swap_role {
-        SwapRole::Alice => State::Alice(AliceState::StartA { local_trade_role }),
```
h4sh3d, line 84: `TradeRole` is `Copy` so we could probably simplify with `*local_trade_role` here.

src/swapd/runtime.rs
```diff
@@ -1593,6 +1602,9 @@ impl Runtime {
                         .unwrap_or_else(|_| Duration::from_secs(0))
                         .as_secs(),
                     public_offer: self.public_offer.clone(),
+                    local_trade_role: self.local_trade_role.clone(),
```
h4sh3d, line 1605: Same here I think (less important as it doesn't add new lines, but I'm sure clippy has something to say on using clone on copy)

h4sh3d (commented):
Minor comments, otherwise ACK

h4sh3d (approved):

PR #770, opened 2022-11-11
Database: Add NodeInfo to CheckpointEntry

TheCharlatan:
Also adds the required members to a swap's checkpoint to allow us to get the required CheckpointEntry members from a a swap checkpoint. 

This is another breaking change to the database schema.
Based on #769


h4sh3d (approved):
ACK, needs rebase

PR #772, opened 2022-11-14
Relisten maker restore

TheCharlatan:
A restored maker swap will now launch a peerd listener. Once a restore transitions into `SwapdRunning` the `expected_counterparty_node_id` is used to identify the correct incoming connection.

### DIFF
--- a/src/farcasterd/trade_state_machine.rs
+++ b/src/farcasterd/trade_state_machine.rs
@@ -1196,10 +1196,7 @@ fn attempt_transition_from_restoring_swapd_to_swapd_running(
         {
             runtime.handle_new_connection(event.source.clone());
 
-            info!(
-                "{} | Peerd connected for restored swap",
-                swap_id.bright_blue_italic()
-            );
+            info!("{} | Peerd connected for restored swap", swap_id.swap_id());
             peerd = Some(event.source.clone());
         }
         (BusMsg::Ctl(CtlMsg::ConnectFailed), source)
@@ -1214,7 +1211,7 @@ fn attempt_transition_from_restoring_swapd_to_swapd_running(
                 if let Some(node_id) = expected_counterparty_node_id {
                     if source.node_addr() == Some(NodeAddr::new(node_id, bind_addr)) {
                         info!(
-                            "{} | Peerd connection for restored swap",
+                            "{} | Peerd connected for restored swap",
                             swap_id.bright_blue_italic()
                         );
                         peerd = Some(source);
--- a/src/farcasterd/trade_state_machine.rs
+++ b/src/farcasterd/trade_state_machine.rs
@@ -635,13 +635,29 @@ fn attempt_transition_to_restoring_swapd(
                 public_offer.offer.accordant_blockchain.try_into()?,
                 public_offer.offer.network,
             )?;
-            let listening =
-                if let Err(err) = runtime.listen(runtime.config.get_bind_addr().unwrap()) {
-                    warn!("failed to re-listen on restore: {}", err);
-                    false
-                } else {
-                    true
-                };
+
+            let listening = if trade_role == TradeRole::Maker {
+                match runtime.config.get_bind_addr() {
+                    Ok(bind_addr) => {
+                        if let Err(err) = runtime.listen(bind_addr) {
+                            warn!("failed to re-listen on restore: {}", err);
+                            false
+                        } else {
+                            true
+                        }
+                    }
+                    Err(err) => {
+                        let msg = format!(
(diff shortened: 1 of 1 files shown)


src/farcasterd/trade_state_machine.rs
```diff
@@ -624,6 +635,14 @@ fn attempt_transition_to_restoring_swapd(
                 public_offer.offer.accordant_blockchain.try_into()?,
                 public_offer.offer.network,
             )?;
+            let listening =
+                if let Err(err) = runtime.listen(runtime.config.get_bind_addr().unwrap()) {
```
h4sh3d, line 639: Handle the `unwrap` here is required IMO, as we are warning if the `listen` fails we should warn the same way if `.get_bind_addr()` fails too, otherwise it is not possible to restart and handle the swap without a connection.

src/farcasterd/trade_state_machine.rs
```diff
@@ -1171,6 +1193,22 @@ fn attempt_transition_from_restoring_swapd_to_swapd_running(
             runtime.handle_failed_connection(event.endpoints, source.clone())?;
             expect_connection = false;
         }
+        (BusMsg::Ctl(CtlMsg::Hello), source) if trade_role == TradeRole::Maker => {
+            if let Some(node_id) = expected_counterparty_node_id {
+                if source.node_addr()
+                    == Some(NodeAddr::new(
+                        node_id,
+                        runtime.config.get_bind_addr().unwrap(),
```
h4sh3d, line 1201: What about making the `get_bind_addr` non-failing. If the config is found but errors we return default values. That would also fix the issue above.

src/farcasterd/trade_state_machine.rs
```diff
@@ -1171,6 +1193,22 @@ fn attempt_transition_from_restoring_swapd_to_swapd_running(
...
+                    ))
+                {
+                    info!(
+                        "{} | Peerd connection for restored swap",
+                        swap_id.bright_blue_italic()
```
h4sh3d, line 1206: Use short form `.swap_id()` helper

src/farcasterd/trade_state_machine.rs
```diff
@@ -1171,6 +1193,22 @@ fn attempt_transition_from_restoring_swapd_to_swapd_running(
...
+        (BusMsg::Ctl(CtlMsg::Hello), source) if trade_role == TradeRole::Maker => {
+            if let Some(node_id) = expected_counterparty_node_id {
+                if source.node_addr()
+                    == Some(NodeAddr::new(
+                        node_id,
+                        runtime.config.get_bind_addr().unwrap(),
+                    ))
+                {
+                    info!(
+                        "{} | Peerd connection for restored swap",
```
h4sh3d, line 1205: Shouldn't this be the same as `"{} | Peerd connected for restored swap"`?

src/farcasterd/trade_state_machine.rs
TheCharlatan: Ok, I'll see if I can handle this configuration error and let the user know of it.

src/farcasterd/trade_state_machine.rs
TheCharlatan: I think it's good to handle this error and let the user know of the broken config.

TheCharlatan: f71102d758af53c12fd843ca2f20eef87bc6701e

src/farcasterd/trade_state_machine.rs
TheCharlatan: 1908a76b08e8dfcb6283e921fb51fcad88c758d7

src/farcasterd/trade_state_machine.rs
TheCharlatan: 1908a76b08e8dfcb6283e921fb51fcad88c758d7

src/farcasterd/trade_state_machine.rs
h4sh3d: Still miss the `.swap_id()` helper, but rest got fixed!

src/farcasterd/trade_state_machine.rs
h4sh3d: Fixed in f71102d758af53c12fd843ca2f20eef87bc6701e

src/farcasterd/trade_state_machine.rs
h4sh3d: 0419507dd2d2b28d67ba64c8a1b7636d355d1ecc

h4sh3d (approved):

PR #773, opened 2022-11-14
Auto-restore checkpoint on startup

TheCharlatan:


src/cli/opts.rs
```diff
@@ -309,6 +316,34 @@ pub enum OfferSelectorParseError {
     Invalid,
 }
 
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Display, From)]
+pub enum CheckpointSelector {
+    #[display("All")]
```
h4sh3d, line 321: This is equivalent to struct level annotation `display(Debug)`

src/cli/opts.rs
```diff
@@ -99,7 +99,15 @@ pub enum Command {
 
     /// Lists saved checkpoints of the swaps
     #[clap(aliases = &["lc"])]
-    ListCheckpoints,
+    ListCheckpoints {
+        #[clap(
+            short,
+            long,
+            default_value = "all",
+            possible_values = &["all", "All", "available", "Available", "available_for_restore"],
```
h4sh3d, line 107: `available-for-restore` seems more common in cli context

src/config.rs
```diff
@@ -118,6 +129,8 @@ impl Default for Config {
 pub struct FarcasterdConfig {
     /// Sets the auto-funding parameters, default to no auto-fund
     pub auto_funding: Option<AutoFundingConfig>,
+    /// Whether checkpoints should be restored manually, or not
+    pub manual_restore: Option<bool>,
```
h4sh3d, line 133: Because the default behavior is restoring checkpoints I believe this option should be `disable_auto_restore` and could find a default value to `false` when generating new config on disk (the `Default` impl on structs).

h4sh3d (commented):
cACK

My main comment is with the option name currently set to `manual_restore` where I think we could find a better name, but not attached to it at a point where this should not be merged as is.

Also updating our example config would be good.

src/config.rs
TheCharlatan: I started out with `disable_auto_restore`, but went with manual restore instead, because I don't like negative options.

src/cli/opts.rs
```diff
@@ -309,6 +316,34 @@ pub enum OfferSelectorParseError {
...
+    #[display("AvailableForRestore")]
+    AvailableForRestore,
+}
+
+impl FromStr for CheckpointSelector {
+    type Err = CheckpointSelectorParseError;
+    fn from_str(input: &str) -> Result<CheckpointSelector, Self::Err> {
+        match input {
+            "all" | "All" => Ok(CheckpointSelector::All),
+            "available" | "Available" | "available_for_restore" => {
```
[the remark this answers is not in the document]
TheCharlatan: ed1fbb0190dc44982d9c0829830121df88294dd2

src/cli/opts.rs
TheCharlatan: ed1fbb0190dc44982d9c0829830121df88294dd2

src/cli/opts.rs
TheCharlatan: 590da5e4a8c9501c44b655891810e3c871f1e212

src/config.rs
Lederstrumpf: > Should attempt again without making this argument optional and making it default true?

I didn't understand what you're referring to with "this argument" over here without flipping default from auto- to manual restore.

but perhaps you were saying this: what's your take on `auto_restore` defaulting to `true`?
I think that's unambiguous to understand & document, at the cost of being inconsistent with the other defaults all being `false`.

TheCharlatan: Sorry, this was complete word salad.
Yes, what I was suggesting is adding `auto_restore` and defaulting it to `true`.

Lederstrumpf: all good - even King Lear enjoyed word salad from time to time (ok fine: a lot) ;)

> Yes, what I was suggesting is adding auto_restore and defaulting it to true.

let's just go with that then

h4sh3d: 60280a1c9864d4da814748de5984f17e54ee764d

h4sh3d (approved):

h4sh3d:
On the taker side we have a race
```
[2022-12-05T12:26:29Z WARN  farcaster_node::farcasterd::trade_state_machine] failed to reconnect to peer on restore: Farcaster node errors: Peer keys not ready yet
```

h4sh3d:
The last set of patches works locally for the taker side, the swaps are correctly auto-restored on node start-up. I re-enabled the feature in tests to see if we break them, but we could just disable it for now to move forward.

TheCharlatan:
Nice work, 
ACK [559cb75](https://github.com/farcaster-project/farcaster-node/pull/773/commits/559cb75e3597eae329e2e9fef71b6006edd8ecc0)

PR #778, opened 2022-11-16
Grpc: Support list offers endpoint

TheCharlatan:
This allows the grpc endpoint to make use of the list offers and their status selector.


h4sh3d (approved):
tACK with the GUI

I left this comment for another PR later:
In general we should try to return more data with the gRPC service, e.g. here returning the status (if easily possible) would allow the UI to build easier contextual actions such as remove the revoke button on passed offers.

PR #779, opened 2022-11-16
Farcaster: Cleanup dangling swap info on restart

TheCharlatan:
If a swap is killed before funding, no checkpoint is written, but the offer status of the involved offer is still set to InProgress. Move the persisted offer status from InProgress to Abort on restart.

### DIFF
--- a/src/databased/runtime.rs
+++ b/src/databased/runtime.rs
@@ -235,8 +235,7 @@ impl Runtime {
                     })
                     .map(|c| c.public_offer)
                     .collect();
-                let dangling_offers: Vec<PublicOffer> = self
-                    .database
+                self.database
                     .get_offers(OfferStatusSelector::InProgress)?
                     .drain(..)
                     .filter_map(|o| {
@@ -246,11 +245,11 @@ impl Runtime {
                             None
                         }
                     })
-                    .collect();
-                for offer in dangling_offers.iter() {
-                    self.database
-                        .set_offer_status(&offer, &OfferStatus::Ended(Outcome::Abort))?;
-                }
+                    .map(|offer| {
+                        self.database
+                            .set_offer_status(&offer, &OfferStatus::Ended(Outcome::Abort))
+                    })
+                    .collect::<Result<_, _>>()?;
             }
 
             _ => {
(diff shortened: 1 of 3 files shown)


src/databased/runtime.rs
```diff
@@ -224,6 +225,34 @@ impl Runtime {
...
+                    .drain(..)
+                    .filter_map(|o| {
+                        if !checkpointed_pub_offers.contains(&o.offer) {
+                            Some(o.offer)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect();
+                for offer in dangling_offers.iter() {
```
h4sh3d, line 250: Style: i think we can avoid collect and creating a new iterator and just have a for_each after the filter_map. Or is it like that because self.database?

h4sh3d (approved):
cACK

src/databased/runtime.rs
TheCharlatan: > Or is it like that because self.database?

It is like that so we can report the Error of `set_offer_status` to the upper level. Is there a better way of doing that?

h4sh3d: Not sure if it is better but we can achieve the same with:
```rust
                self.database
                    .get_offers(OfferStatusSelector::InProgress)?
                    .drain(..)
                    .filter_map(|o| {
                        if !checkpointed_pub_offers.contains(&o.offer) {
                            Some(o.offer)
                        } else {
                            None
                        }
                    })
                    .map(|offer| {
                        self.database
                            .set_offer_status(&offer, &OfferStatus::Ended(Outcome::Abort))
                    })
                    .collect::<Result<_, _>>()?;
```
Collect can transform a `Vec<Res<>>` into `Res<Vec<>>`, this should fail on the first error.

TheCharlatan: I found this as well in the meantime from rust by example: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

I think this is indeed preferable, will apply this pattern elsewhere when writing new code.

TheCharlatan: 635aa9af9badaa1b466067f24d568e4d8d4d9b72

Issue #780, opened 2022-11-16
Add block height and swap id to persisted funding addresses


TheCharlatan:
So we know from which height to sweep the Monero funding address and to which swap they belonged to when cleaning up dangling swap data.

PR #782, opened 2022-11-16
Database: Change secret key encoding

TheCharlatan:
Bitcoin addresses now additionally store an optional swap_id.

Monero addresses additionally store an optional swap_id and optional
creation height.

Also add cleaner database encoding error handling.

This allows us to know from which height to sweep the Monero funding address and to which swap the funding addresses belong when cleaning up dangling swap data.

### DIFF
--- a/src/bus/types.rs
+++ b/src/bus/types.rs
@@ -30,14 +30,29 @@ pub enum AddressSecretKey {
     #[display("addr_key({address}, ..)")]
     Bitcoin {
         address: bitcoin::Address,
-        secret_key: bitcoin::secp256k1::SecretKey,
+        secret_key_info: BitcoinSecretKeyInfo,
     },
     #[display("addr_key({address}, ..)")]
     Monero {
         address: monero::Address,
-        view: monero::PrivateKey,
-        spend: monero::PrivateKey,
+        secret_key_info: MoneroSecretKeyInfo,
     },
+}
+
+#[derive(Clone, Debug, Display, Eq, PartialEq, NetworkDecode, NetworkEncode)]
+#[display("bitcoin secret key info")]
+pub struct BitcoinSecretKeyInfo {
+    pub swap_id: Option<SwapId>,
+    pub secret_key: bitcoin::secp256k1::SecretKey,
+}
+
+#[derive(Clone, Debug, Display, Eq, PartialEq, NetworkDecode, NetworkEncode)]
+#[display("monero secret key info")]
+pub struct MoneroSecretKeyInfo {
+    pub swap_id: Option<SwapId>,
+    pub view: monero::PrivateKey,
+    pub spend: monero::PrivateKey,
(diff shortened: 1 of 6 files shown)


Lederstrumpf:
dumb q: I've forgotten why we wanted this: Mind adding the context again, either here, in issue #780, or as comment in the code?

iirc it was to avoid scanning for funding txs too far when restoring, but then I don't understand why only the monero addresses have a creation height.

TheCharlatan:
> dumb q: I've forgotten why we wanted this: Mind adding the context again, either here, in issue #780, or as comment in the code?

Yes, added a one-sentence description to each. It boils down to a) persisting when the Monero address was created so we know from which height onwards to scan and b) when cleaning up dangling swap data to identity swaps that already entered the funding stage.

> iirc it was to avoid scanning for funding txs too far when restoring, but then I don't understand why only the monero addresses have a creation height.

The creation height for the address only matters for Monero, Bitcoin does not care about blockchain depth.

src/bus/types.rs
```diff
@@ -30,16 +30,31 @@ pub enum AddressSecretKey {
...
-        view: monero::PrivateKey,
-        spend: monero::PrivateKey,
+        secret_key_info: MoneroSecretKeyInfo,
     },
 }
 
+#[derive(Clone, Debug, Display, Eq, PartialEq, NetworkDecode, NetworkEncode)]
+#[display("bitcoin secret key info")]
+pub struct BitcoinSecretKeyInfo {
+    pub swap_id: Option<SwapId>,
```
Lederstrumpf, line 45: I don't understand why this is an option, and not just the naked type? We only instantiate secret keys for both sides in the context of a swap. The only case where this ain't `Some(..)` is inside the [lmdb test](https://github.com/TheCharlatan/farcaster-node/blob/cad412a782ee6755481c538c003a3f901b8c0db9/src/databased/runtime.rs#L692).

TheCharlatan: I made this an Option for flexibility purposes, we might have a secret key in the future that is not attached to a swap id. It's a _very_ speculative reason, so if that's not good enough for you, let me know.

h4sh3d: Then it's equivalent to some _NULLABLE_ column in other type of db, which is fine by me if it helps us avoid breaking the storage format later. I'd just add a comment on the struct to avoid confusion in the future (not related to protocol, just storage layout)

Lederstrumpf: thanks for the rationale - I'm fine to keep the speculative flexibility to drop swap_id coupling: I have a bunch of checkpoints that are useless at this stage (unless I go through the pain of rebuilding the old fc versions, or even just figuring out which ones those were), so let's not risk more of that for users ;)

Lederstrumpf (approved):
tACK @ 46221f510d84b4f731e9c5b60a73d3797a9220c8 (has other PRs on it too).

PR #761, opened 2022-11-08
Add richer data to the progress messages

TheCharlatan:
There are now three possible progress messages that a swap can issue. I message is just a string containing information. A transition goes from an old state to a new state. A state update updates the current state. Farcasterd only keeps the latest state update around, any previous state updates are discarded.

The progress now serves additional information about the confirmation status of various transactions during a swap. Additionally, it provides the block height of a running swap daemon.

I can be pretty flexible with the provided information here, so please let me know what additions / changes / re-designs you want.

Sample Output:

```
> swap1-cli progress 0xaf373e2400556909ab5b90959242979ac6bb05e0c48888d24533095fc7ead6ab                          

progress:
  - message: Proposing to take swap 0xbad3…3565 to Maker remote peer
  - transition:
      old_state: StartA
      new_state: CommitA
  - transition:
      old_state: CommitA
      new_state: RevealA
  - transition:
      old_state: RevealA
      new_state:
        RefundSigA:
          arb_block_height: 2405384
          acc_block_height: 1219189
          arb_locked: false
          acc_locked: false
          buy_published: false
          cancel_seen: false
          refund_seen: false
          overfunded: false
          arb_lock_confirmations: ~
          acc_lock_confirmations: ~
          blocks_until_cancel_possible: ~
          cancel_confirmations: ~
          blocks_until_punish_possible: ~
          blocks_until_safe_buy: ~
  - StateUpdate:
      RefundSigA:
        arb_block_height: 2405386
        acc_block_height: 1219192
        arb_locked: true
        acc_locked: true
        buy_published: true
        cancel_seen: false
        refund_seen: false
        overfunded: false
        arb_lock_confirmations: 2
        acc_lock_confirmations: 1
        blocks_until_cancel_possible: 2
        cancel_confirmations: ~
        blocks_until_punish_possible: ~
        blocks_until_safe_buy: 0
  - transition:
      old_state:
        RefundSigA:
          arb_block_height: 2405386
          acc_block_height: 1219192
          arb_locked: true
          acc_locked: true
          buy_published: true
          cancel_seen: false
          refund_seen: false
          overfunded: false
          arb_lock_confirmations: 2
          acc_lock_confirmations: 1
          blocks_until_cancel_possible: 2
          cancel_confirmations: ~
          blocks_until_punish_possible: ~
          blocks_until_safe_buy: 0
      new_state:
        FinishA: SuccessSwap
```

PR #783, opened 2022-11-17
Swapd: Some patches and improvements

TheCharlatan:
Based on #761

Correct log message for when monero reaches the buyer.
Log incoming monero syncer watch addr and tx tasks on debug.
Record and checkpoint the monero address creation height.
Improve formatting of two log messages.
Swapd: Safer behaviour around publishing transaction. Always publish the refund transaction if you can, inform the user if punish was possible in the meantime

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -737,7 +737,7 @@ impl Runtime {
             {
                 // checkpoint swap pre lock bob
                 debug!("{} | checkpointing bob pre lock state", self.swap_id);
-                // Set the monero address creation height for Bob to before setting the first checkpoint
+                // Set the monero address creation height for Bob before setting the first checkpoint
                 if self.monero_address_creation_height.is_none() {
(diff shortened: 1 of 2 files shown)


src/swapd/runtime.rs
```diff
@@ -733,6 +737,11 @@ impl Runtime {
             {
                 // checkpoint swap pre lock bob
                 debug!("{} | checkpointing bob pre lock state", self.swap_id);
+                // Set the monero address creation height for Bob to before setting the first checkpoint
```
h4sh3d, line 740: typo

h4sh3d (commented):
ACK

src/swapd/runtime.rs
TheCharlatan: f59932b8b9e1c0c9fdd14d193a114d5731907af7

h4sh3d (approved):

PR #784, opened 2022-11-18
Feat: Use creation height when sweeping Monero

TheCharlatan:
This allows us to use the recorded creation height of a Monero address when sweeping from that address.

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -1566,13 +1566,10 @@ impl Runtime {
                             .pop()
                             .unwrap();
                         if let (
-                            BusMsg::Sync(SyncMsg::Task(Task::SweepAddress(mut task))),
+                            BusMsg::Sync(SyncMsg::Task(Task::SweepAddress(task))),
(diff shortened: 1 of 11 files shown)


src/swapd/runtime.rs
```diff
@@ -1460,6 +1461,25 @@ impl Runtime {
...
+            CtlMsg::SetAddressSecretKey(address_secret_key) => {
+                if let AddressSecretKey::Monero {
+                    address,
+                    mut secret_key_info,
+                } = address_secret_key
+                {
+                    secret_key_info.creation_height = self.monero_address_creation_height;
+                    endpoints.send_to(
+                        ServiceBus::Ctl,
+                        ServiceId::Wallet,
```
h4sh3d, line 1473: Are we forced to impersonate wallet here? For me it seems a bad practice if we are not forced to.

TheCharlatan: No longer relevant, since the wallet logic now lives in the swap daemon.

PR #787, opened 2022-11-18
Peerd connect treatmeant

TheCharlatan:
Based on @Lederstrumpf farcasterd.dev-testing branch.


src/farcasterd/trade_state_machine.rs
```diff
@@ -270,7 +302,40 @@ impl TradeStateMachine {
...
+            TradeStateMachine::SwapdRunning(SwapdRunning {
+                peerd,
+                public_offer,
+                ..
+            }) => {
+                // Peerd is none, implies a connect event is possible. A connect
+                // is only possible to a listening peer, its address thus has to
+                // match the address of the offer.
+                if peerd.is_none() {
+                    Some(ServiceId::Peer(node_addr_from_public_offer(public_offer)))
```
Lederstrumpf, line 334: with a taker that successively takes offers from the maker once the last completes: if the last successful run was Bob maker, then Alice taker will most likely terminate connection before Bob has sweeped his xmr. Then Alice launches another swap.

In my understanding, a new peer connection will not be registered because `awaiting_connect_from(..)` will still match against the tsm of the last successful run here, since Bob's peerd is down, yet `node_addr_from_public_offer(..)` is the same for the old and the new swap?

TheCharlatan: I still don't understand quite understand how this triggers a state where the `peerd.is_none()`, but this has to be the case, so I added an additional check to only trigger this when case when taking / connecting a swap.

src/farcasterd/trade_state_machine.rs
```diff
@@ -768,7 +947,11 @@ fn attempt_transition_from_swapd_launched_to_swapd_running(
...
         _ => {
-            trace!("BusMsg {} invalid for state swapd launched", event.request);
+            trace!(
+                "{} | BusMsg {} invalid for state swapd launched",
+                swap_id,
+                event.request
+            );
         }
     }
     let peerd_up = runtime.registered_services.contains(&peerd);
```
Lederstrumpf, line 957: with a unique `ServiceId` for every peerd instantiation, if we create a new peerd instance for a remote and kill the old one, this will return `false` since the old peerd's `ServiceId` is cached in the tsm but won't be in the `runtime.registered_services` anymore, so swap will never launch. So we can instead here check only against the node address of the peerd key in `runtime.registered_services`, but if there are multiple concurrent peerds living for said remote, then still should ensure that we launch with the latest one.

Lederstrumpf: I've pushed a commit that does an ugly impl of this. TODO if we go with this change:
1. it currently just uses the first peerd it finds in registered services - imo we'd need to encode creation time in the peerd `ServiceId` to do that correctly
2. the `SwapdLaunched` struct should not keep the full peerd `ServiceId` since only the `NodeAddr` is read with this change

TheCharlatan:
Closing this PR in favor of #792

PR #789, opened 2022-11-26
Swap wallet

TheCharlatan:
The goal of this PR is moving most of the swap related logic from walletd to swapd. The remaining logic in walletd just instantiates keys and a key manager that can be used by the swap daemons. 

When a swap is created, farcasterd requests a new key manager from walletd and passes it to a spawned swapd.

Logic that previously lived in walletd is now called directly as a function from swapd. This greatly simplifies the flow of logic inside swapd.


src/swapd/runtime.rs
```diff
@@ -612,46 +552,42 @@ impl Runtime {
                         )?;
                     }
                 }
-                // forward the message to wallet to trigger creation of reveal message
-                debug!("{} | forward maker commitment to wallet", self.swap_id);
-                self.send_wallet(
```
h4sh3d, line 617: As `send_wallet` is now unused we can remove it from `service.rs`, I linked the related issue already.

src/walletd/runtime.rs
```diff
@@ -135,1192 +64,33 @@ impl esb::Handler<ServiceBus> for Runtime {
...
-                        ))),
-                    )?;
-                } else {
-                    error!("Call to refund transaction expects an Alice wallet");
+            CtlMsg::CreateSwapKeys(public_offer, wallet_token) => {
+                if wallet_token != self.wallet_token {
+                    return Err(Error::InvalidToken);
                 }
+                let wallet_index = self.node_secrets.increment_wallet_counter();
+                let key_manager = KeyManager::new(self.node_secrets.wallet_seed, wallet_index)?;
```
h4sh3d, line 83: Do we have access to the swap id here if we need to? I'd like to refactor the `KeyManager` to adjust this new usage (with the cache system) and more importantly to take the swap id instead of a counter (which is super insecure). We also can make use of the offer uuid but I'd prefer the swap id.

src/swapd/swap_state.rs
```diff
@@ -548,6 +556,35 @@ impl State {
             _ => unreachable!("checked state"),
         }
     }
+
+    /// Update Bob Reveal with Alice's reveal
+    pub fn b_sup_alice_reveal(&mut self, reveal: Reveal) -> bool {
```
h4sh3d, line 561: I'm asking here because otherwise I don't ask ;p

What's the meaning of `.._sup_..` in the function name? I get the `a|b_` for Alice and Bob but not the sup.

TheCharlatan: I don't know either

src/walletd/runtime.rs
h4sh3d: I didn't check since a long time ago, but this counter is not persisted right, it resets at every boot? The seed is persisted which is correct, so if we use the same counter twice we generate the same keys.

TheCharlatan: No, the counter is persisted as well.

h4sh3d: Ok, the counter is persisted on disk at each increment. Then you can ignore my comment.

But if we want to simplify this process and do not need to persist changes on disk every time we can remove the counter and use the swap id at manager instantiation.

h4sh3d (commented):
I have to say that I'm genuinely impressed by the "low complexity" of this refactor. Very nice work!

This is my first feedback, all are minor suggestion that can be freely dropped if considered useless or out-of-scope

src/farcasterd/trade_state_machine.rs
```diff
@@ -475,8 +494,8 @@ fn attempt_transition_to_taker_connect_or_take_offer(
     match event.request.clone() {
         BusMsg::Ctl(CtlMsg::TakeOffer(PubOffer {
             public_offer,
-            external_address,
-            internal_address,
+            bitcoin_address: external_address,
+            monero_address: internal_address,
```
Lederstrumpf, line 498: I'd get rid of renaming to external/internal address here entirely since
1. imo, those names were conceptually misleading to begin with
2. they're just being renamed to `arb_addr`/`acc_addr` thereafter anyway

src/swapd/runtime.rs
```diff
@@ -462,82 +472,20 @@ impl Runtime {
...
-            debug!(
-                "{} | trying checkpoint Bob pre buy but Buy Procedure Signature not received yet",
-                self.swap_id,
-            );
-        }
-        Ok(())
-    }
-
-    fn handle_msg(
+    pub fn handle_msg(
```
Lederstrumpf, line 475: assuming was only there for iterating fast and can be reverted
```suggestion
    fn handle_msg(
```

src/swapd/runtime.rs
```diff
@@ -551,15 +499,6 @@ impl Runtime {
...
         }
 
-        // check if message swap id matches internal swap id
-        if request.swap_id() != self.swap_id() {
-            return Err(Error::Farcaster(format!(
-                "Incorrect swap id: expected {}, found {}",
-                self.swap_id(),
-                request.swap_id(),
-            )));
-        }
```
Lederstrumpf, line 561: why can we be sure that this can't happen?

Lederstrumpf (commented):
first feedback:

> I have to say that I'm genuinely impressed by the "low complexity" of this refactor. Very nice work!

fully agree: this gets rid of some twisted logic really cleanly. great job, can't wait to see it shape up with #804!

src/swapd/runtime.rs
TheCharlatan: [f3dfc10](https://github.com/farcaster-project/farcaster-node/pull/789/commits/f3dfc105803216777d7857558f05a7ae94c9f439)

src/walletd/runtime.rs
TheCharlatan: I'd rather have a deterministic seed than not. If there is ever a case where a user looses the funding address data, he could still re-create it from the wallet secret and the counter.

src/swapd/runtime.rs
TheCharlatan: I removed it, because it was a code duplicate.

src/databased/runtime.rs
```diff
@@ -84,27 +84,18 @@ impl Runtime {
...
-
-                        debug!("setting swap checkpoint");
-                    }
+                let info = CheckpointEntry {
+                    swap_id,
+                    public_offer: state.public_offer.clone(),
+                    trade_role: state.local_trade_role.clone(),
+                    expected_counterparty_node_id: state.connected_counterparty_node_id.clone(),
                 };
+                debug!("setting checkpoint info entry");
```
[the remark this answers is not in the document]
TheCharlatan: 37d1eb609589bc6ece3c7bae4c71a5e127532307

src/farcasterd/trade_state_machine.rs
```diff
@@ -173,6 +178,20 @@ pub struct SwapdRunning {
     trade_role: TradeRole,
 }
 
+pub enum ConsumedOfferRole {
+    Taker,
+    Maker(Commit),
+}
+
+impl ConsumedOfferRole {
+    pub fn trade_role(&self) -> TradeRole {
```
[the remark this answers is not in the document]
TheCharlatan: 1a2b9097e459b4e238fa5323ce688091feef5cc6

src/bus/ctl.rs
```diff
@@ -212,12 +223,30 @@ pub struct GetKeys(pub Token);
 pub struct LaunchSwap {
     pub local_trade_role: TradeRole,
     pub public_offer: PublicOffer,
-    pub local_params: Params,
     pub swap_id: SwapId,
-    pub remote_commit: Option<Commit>,
-    pub funding_address: Option<bitcoin::Address>,
```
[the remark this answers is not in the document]
TheCharlatan: c2d50d1dbe484e2ebc9e3bb5925cebaf33bd377f

src/swapd/runtime.rs
TheCharlatan: 25d2e10f325e6b5bf188d616c8b85ea7d242b4cb

src/swapd/runtime.rs
```diff
@@ -612,46 +552,42 @@ impl Runtime {
...
+            //  3. Send the Alice Reveal message to the counterparty
             //
             // Upon reception Bob needs to
             //  1. Validate the parameters
-            //  2. Add the reveal message to the pending message for later use
-            //  3. Send the funding information message to farcaster
-            //  4. Watch the arbitrating funding address if we are maker
+            //  2. IF the fee estimation is completed, add the reveal message to the state
+            //  2.3. Send the funding information message to farcaster
+            //  2.4. Watch the arbitrating funding address if we are maker
```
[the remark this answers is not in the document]
TheCharlatan: 3fb891d4a89abd092f75003b2ca9ab56c8d23b12

src/walletd/runtime.rs
h4sh3d: What I propose is: seed persited on disk + swap id to generate the keys (instead of a counter). The user has no knowledge about the counter's value used, but he knows the swap id.

src/swapd/runtime.rs
```diff
@@ -612,46 +552,42 @@ impl Runtime {
...
-                    && self.state.remote_commit().is_some() =>
-            {
-                // forward message to peer
-                debug!("{} | send reveal peer message to peerd", self.swap_id);
+                let reveal = self
+                    .wallet
+                    .as_mut()
+                    .expect("should be initialized")
+                    .handle_maker_commit(remote_commit, self.swap_id.clone())?;
+                info!(
```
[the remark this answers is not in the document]
TheCharlatan: 5f64790ff47d7d7bafac1b0cb8ff32a399553e93

src/walletd/runtime.rs
TheCharlatan: This would mean that the user would have to keep track of all swap ids in perpetuity. For the counter, the user just needs the current count (a single integer) to generate all the keys, which is currently persisted in the same file as the secret.

So the current situation is:

key.dat: NodeSecrets { peerd_secret_key, wallet_seed, wallet_counter }

And here is what I think it will look like if we use the swap id instead:

key.dat: Node Secrets { peerd_secret_key, wallet_seed, Vec<SwapId> }

I'd prefer the first option, seems much safer to me.

TheCharlatan:
I see your address naming convention comments and I hear you. In this PR, I just want to get rid of internal/external, because it really does not make any sense. If it is fine by I'd propose tackling the rename in another PR.

src/walletd/runtime.rs
h4sh3d: The wallet counter is incremented/updated for each new swap, and the last generated `u32` is persisted in the file.

https://github.com/farcaster-project/farcaster-node/blob/c92187fc65a561174b401619fc43b74f7e738d2d/src/walletd/runtime.rs#L299-L301

https://github.com/farcaster-project/farcaster-node/blob/c92187fc65a561174b401619fc43b74f7e738d2d/src/walletd/opts.rs#L148

If you do 10 swaps and you need the keys for the `#3`, if you know the correct order you can load the counter and go ask for the correct index. For me this seems harder than having access the the swap id.

I wouldn't store the swap ids in the file, just the seed. Swap id can be in databased if wanted (the history of all swaps) or the responsibility of the user.

Btw I'm fine with the current design, but I'd prefer using static seed + swap id instead of static seed + counter. I don't see strong security difference between the two, only UX.

diagrams/sequencediagram.txt
```diff
@@ -62,49 +60,32 @@ m_farcasterd<-m_syncer:Ctl Hello [from Syncer Bitcoin (if newly launched) OR Swa
...
-m_wallet -> m_swap : Msg Reveal (maker is sender)
 peerd <- m_swap : Msg Reveal (maker is sender)
 peerd -> t_swap : Msg Reveal (maker is sender)
-group for Alice swap role
-t_swap -> t_wallet : Msg Reveal (maker is sender)
-end
+
 group for Bob swap role
-t_swap -> t_swap : ADD PENDING Msg Reveal
+t_swap -> t_swap : ADD Msg CoreArb
```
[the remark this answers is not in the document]
TheCharlatan: I'm not sure where you are seeing CoreArb as pending? But CoreArb is still wrong, it adds the Reveal message to the state.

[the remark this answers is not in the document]
TheCharlatan: 2d0938a51c50f50dbef8b49e40de7b9078ab9f54

src/walletd/runtime.rs
TheCharlatan: If you have still have access to the swap database, you will still have access to its corresponding keys anyway. So there is no point in using the swap id with the seed to recover anything, since you already have the data. If you use just the counter you still have redundancy in case something happens to the database.

On a different note: Didn't we decide to phase out the swap id anyway? I'm really having a hard time seeing the benefits here.

diagrams/sequencediagram.txt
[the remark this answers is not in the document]
Lederstrumpf: ah yeah - I assumed that because there was no CoreArb sent by the counterparty, this was the taker already creating CoreArb to send later (pending), but that was mistaken and it now makes perfect sense again, thanks!

src/swapd/runtime.rs
Lederstrumpf: oops, of course 🙈

src/farcasterd/trade_state_machine.rs
TheCharlatan: ffde2b9be5584b15d1ba44ffbaf8a7153cf8e1b0

src/farcasterd/trade_state_machine.rs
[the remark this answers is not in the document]
h4sh3d: ...not sure it was a good indication from me as `From` consumes the value et we'd like to avoid that. This is minor optimization tho, I'm fine keeping the patch or removing it. Sorry for the noise

h4sh3d (approved):
tACK ffde2b9be5584b15d1ba44ffbaf8a7153cf8e1b0

PR #796, opened 2022-12-01
Farcaster: Handle take offer failure

TheCharlatan:
The maker now replies to the taker with  the `OfferNotFound` message in case the offer is no longer available. The taker, on receipt, then aborts the already spawned swap.

### DIFF
--- a/src/farcasterd/runtime.rs
+++ b/src/farcasterd/runtime.rs
@@ -826,8 +826,8 @@ impl Runtime {
 
     fn match_request_to_syncer_state_machine(
         &mut self,
-        req: BusMsg,
-        source: ServiceId,
+        req: &BusMsg,
+        source: &ServiceId,
     ) -> Result<Option<SyncerStateMachine>, Error> {
         match (req, source) {
             (BusMsg::Ctl(CtlMsg::SweepAddress(..)), _) => Ok(Some(SyncerStateMachine::Start)),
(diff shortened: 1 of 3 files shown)


src/farcasterd/runtime.rs
```diff
@@ -913,18 +913,14 @@ impl Runtime {
         source: ServiceId,
         endpoints: &mut Endpoints,
     ) -> Result<(), Error> {
-        if let Some(tsm) =
-            self.match_request_to_trade_state_machine(request.clone(), source.clone())?
-        {
+        if let Some(tsm) = self.match_request_to_trade_state_machine(&request, &source)? {
```
h4sh3d, line 916: I like this!

src/farcasterd/runtime.rs
```diff
@@ -961,6 +957,16 @@ impl Runtime {
                     )?;
                     Ok(())
                 }
+                BusMsg::P2p(PeerMsg::TakerCommit(TakerCommit { commit, .. })) => {
+                    debug!("{} | Offer already taken or aborted, replying with offer not found to the counterparty", commit.swap_id());
```
h4sh3d, line 961: Maybe log this as info with the offer uuid. I guess it's an interesting info to have as a maker

h4sh3d (approved):
cACK

This could also be deployed on farcaster.dev IMO

src/farcasterd/runtime.rs
h4sh3d: 0f8e1fc5195c270c73b402e33e388cc9eba85e4a

h4sh3d: This is done for taker for now

PR #800, opened 2022-12-02
Config: expose swap finality parameters

h4sh3d:
This PR introduce a new section`[swap...]` in the config file `farcasterd.toml`. The swap section, for now, contains finality and safety parameters per chain per network. A chain can accept `arbitrating` parameters or `accordant` parameters, currently we only support btc and xmr:
https://github.com/farcaster-project/farcaster-node/blob/981c0d1d93986f7584bd4b32efbbe41e5ae1854e/src/config.rs#L239-L246

The configuration can then return a parsed swap configuration for a given network, arb chain, and acc chain
https://github.com/farcaster-project/farcaster-node/blob/981c0d1d93986f7584bd4b32efbbe41e5ae1854e/src/config.rs#L150-L156

We have default values for mainnent and testnet for Bitcoin and Monero, if no config is found in the file we return those default value (backward compatibility), otherwise an error is returned. The configuration translate the real chain (Btc, Xmr) into its interface (Arb, Acc) as the swap is started with the offer that contains the chains used.

Launching a swap now takes the parsed swap config to 
https://github.com/farcaster-project/farcaster-node/blob/981c0d1d93986f7584bd4b32efbbe41e5ae1854e/src/farcasterd/runtime.rs#L1244-L1260

Cli arguments of swapd have been reworked to accept the new arguments. (Breaking change with other bins that use swapd directly)

The new `chains.rs` file contains enums for listing support arb/acc chains in our node (e.g. btc as acc is in theory possible but probably not supported in the current architecture, but that would be a nice test to see where the current design falls apart). These structs can be reused later elsewhere (e.g. make cli args) to ensure we accept only valid params.


h4sh3d:
@Lederstrumpf @TheCharlatan do we want to expose different finality and safety margin per network? (And are those names good?)

TheCharlatan:
names lgtm

farcasterd.toml
```diff
@@ -37,6 +37,23 @@ bitcoin_cookie_path = "~/.bitcoin/testnet3/.cookie"
...
+[swap.bitcoin]
+# Avoid broadcasting a transaction if a race can happen in # blocks
+#
+# E.g. if the safety margin is set to 3 and the the cancel transaction becomes
+# broadcastable in 4 blocks we are safe and we do publish buy, if it becomes
+# broadcastable in 3 blocks or less we do not publish the buy as the safety
+# margin is not respected.
+safety_margin = 3
+# Number of confirmations required to consider a transaction final
+finality = 1
```
Lederstrumpf, line 50: nit: for consistency, I'd either
1. `safety_margin` → `safety` or
2. `finality` → `finality_thr`/`finality_threshold` for consistency.

Lederstrumpf:
> @Lederstrumpf @TheCharlatan do we want to expose different finality and safety margin per network? (And are those names good?)

agree it makes sense to have different finality thresholds per network. for safety, it would be more consistent to call this `safety_threshold` rather than `safety_margin`, but it works in the opposite direction (upper vs lower bound) to finality, so having distinct names may avoid someone misinterpreting the direction.
I'll also put suggestions of `finality_min` and `safety_max` on the table to be absolutely clear on direction since they require less thinking / a lower command of English to be understood unambiguously.

farcasterd.toml
h4sh3d: What I initially disliked with `threshold` is we can put it everywhere. For now the variant I like the most is renaming `safety` and keeping `finality`. IMO the comments are very important, good descriptions are also key for a good config file.

src/swapd/temporal_safety.rs
```diff
@@ -2,23 +2,40 @@ use crate::Error;
 use farcaster_core::blockchain::Blockchain;
 use strict_encoding::{StrictDecode, StrictEncode};
 
+/// Represent a blockchain height
 pub type BlockHeight = u32;
 
+/// Represent a block length or a block number
+pub type BlockSpan = u32;
```
Lederstrumpf, line 9: 👍

src/config.rs
```diff
@@ -226,25 +384,44 @@ pub struct SyncerServers {
     pub monero_wallet_dir: Option<String>,
 }
 
+#[derive(Deserialize, Serialize, Default, Debug, Clone)]
+#[serde(crate = "serde_crate")]
+pub struct Networked<T> {
```
Lederstrumpf, line 389: I both hate & love this struct naming xD
I'm guessing you wanted to avoid Network because this here is limited to config? Then might go for `NetworkConfig`? Anyway, keen to get your rationale for this one on Monday.

Lederstrumpf (approved):
ACK @ bda0705f2404303f052e42a062b59a34399dd3f1
and I look forward to `chains.rs` spreading its claws through the node to make it chain generic  🔪  :D!

Lederstrumpf:
(I'm holding off on merge to get #789 in first since they conflict slightly)

src/config.rs
h4sh3d: :) yes network is to generic, just wanted to express that it’s a wrapper to “network” a type. NetworkConfig would be fine (if we don’t introduce some [network..] section later) or PerNetwork maybe.

Lederstrumpf:
> This PR introduce a new section`[swap...]` in the config file `farcasterd.toml`. The swap section, for now, contains finality and safety parameters per chain per network. A chain can accept `arbitrating` parameters or `accordant` parameters, currently we only support btc and xmr:
> 
> ...

ok - merging :). And nice new PR docs, @h4sh3d!

PR #804, opened 2022-12-04
Swap state machine

TheCharlatan:
The implemented state machine is implemented and documented in the swap state.

For reviewing this PR, I'd suggest following the full paths in the state machine for each Alice and Bob as Maker and Taker on their happy, cancel and abort paths.

There is one function, `execute_state_machine`, with which I am not happy with. It contains a `clone` of the incoming request, and adds a bunch of logic to ensure the prior, unhandled messages are processed when the state machine allows it. Any suggestions for its improvement are very welcome.

WIth the implementation of this PR, I am now also more confident of eventually being able to pause its happy path execution, or manually triggering the cancel path.


src/swapd/runtime.rs
```diff
@@ -1286,7 +424,7 @@ impl Runtime {
                     BusMsg::Sync(SyncMsg::Task(watch_height_bitcoin)),
                 )?;
 
-                trace!("Watch height monero");
+                self.log_trace("Watch height monero");
```
h4sh3d, line 427: General comment about these log trace, isn't the trace level too low? IMO these messages could be at the debug level and trace is only used when printing array of bytes, big hex strings, etc.

h4sh3d: Resolved by async discussion with @TheCharlatan

TheCharlatan (commented):
ACK 693c21583f01cd7992da70013b41aca17eec2cb7

Lederstrumpf (changes requested):
tACK @a5ebfe5c8cce9e65d1ff4bc7f8bfa1c3b85d4aea

This is an *excellent* refactor and was a joy to review, kudos @TheCharlatan! It's turned a lot of logic that was previously wired in a twisted mess sequential and clear: the protocol screams its identity from the code now and can be reasoned about. I've opened up issues #828, #833, #834, and #835 to track issues I ran into during the review that make more sense to address in follow-up PRs than to integrate here.

Stats:
``` bash
[2022-12-12T04:04:23Z INFO  farcaster_node::farcasterd::stats] Swapped(4630) | Refunded(35) / Punished(0) | Aborted(0) | Initialized(4688) / AwaitingFundingXMR(0) / AwaitingFundingBTC(0) / FundedXMR(2332) / FundedBTC(2316) / FundingCanceledXMR(28) / FundingCanceledBTC(0)
```

Lederstrumpf (approved):
tACK @a5ebfe5c8cce9e65d1ff4bc7f8bfa1c3b85d4aea

This is an *excellent* refactor and was a joy to review, kudos @TheCharlatan! It's turned a lot of logic that was previously wired in a twisted mess sequential and clear: the protocol screams its identity from the code now and can be reasoned about. I've opened up issues #828, #833, #834, and #835 to track issues I ran into during the review that make more sense to address in follow-up PRs than to integrate here.

Stats:
``` bash
[2022-12-12T04:04:23Z INFO  farcaster_node::farcasterd::stats] Swapped(4630) | Refunded(35) / Punished(0) | Aborted(0) | Initialized(4688) / AwaitingFundingXMR(0) / AwaitingFundingBTC(0) / FundedXMR(2332) / FundedBTC(2316) / FundingCanceledXMR(28) / FundingCanceledBTC(0)
```

h4sh3d (approved):

Issue #812, opened 2022-12-09
grpcd connects "as peer" to swapd if relaunched manually


Lederstrumpf:
after relaunching manually using `grpcd -d ~/.farcaster -m {data_dir}/msg -x {data_dir}/ctl -i {data_dir}/info -S {data_dir}/sync --grpc-port 50051 --grpc-ip 0.0.0.0`:

```
[2022-12-09T15:31:59Z INFO  farcaster_node::farcasterd::runtime] Service grpcd is now connected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x3923…21ac | Peer grpcd reconnected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x25dd…80ae | Peer grpcd reconnected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x4fed…7676 | Peer grpcd reconnected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x5334…05de | Peer grpcd reconnected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x65fc…b333 | Peer grpcd reconnected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x69cd…7ca6 | Peer grpcd reconnected
[2022-12-09T15:31:59Z INFO  farcaster_node::swapd::runtime] 0x94ec…1e71 | Peer grpcd reconnected
```

PR #813, opened 2022-12-09
TSM: Check if Hello is coming from peerd

TheCharlatan:
Fixes an issue where the reconnect message would be triggered by a Hello of an arbitrary peer.


TheCharlatan:
Don't forget to change the sequence diagram. Make it clearer what happens when state is added.

Lederstrumpf (approved):
ACK @ 37599a6a4e967b7b7744f5fc72b37526da8e49bf. Can't reproduce the issue I was facing in #812 anymore (requires a particularly bad networking day, even for our hetzner nodes :D), but the logic here should certainly catch it.

PR #816, opened 2022-12-12
Grpc: Handle recoverable server runtime panics

TheCharlatan:
If there is an error within the grpc server runtime, it will destroy the server and send a notification to the zmq runtime. The zmq runtime then restarts the grpc server.


h4sh3d (approved):
tACK 0c451b0191f02f2aed1a697ef3d05765c421917a

Panics where very sparse before this PR, more loads on the gRPC server with the GUI will validate the new architecture but for now this seems very solid.

PR #818, opened 2022-12-12
Grpc: Add network selectors to data queries

TheCharlatan:
This allows the grpc server to filter the response contents by certain networks. When calling ListOffers, Checkpoints, FundingAddresses, and NeedsFunding the response is now filtered by the network.


src/grpcd/runtime.rs
```diff
@@ -324,17 +336,17 @@ impl From<StateReport> for farcaster::State {
     }
 }
 
-impl From<farcaster::HealthCheckSelector> for HealthCheckSelector {
-    fn from(s: farcaster::HealthCheckSelector) -> Self {
+impl From<farcaster::NetworkSelector> for HealthCheckSelector {
```
h4sh3d, line 339: Do we still need an `HealthCheckSelector` in `bus` module or should we also match the new gRPC structure?

src/grpcd/runtime.rs
```diff
@@ -1053,6 +1112,22 @@ impl Farcaster for FarcasterService {
                     funding_infos: infos
                         .swaps_need_funding
                         .iter()
+                        .filter(|info| {
+                            if network_selector == NetworkSelector::AllNetworks {
+                                true
+                            } else {
+                                match info {
```
h4sh3d, line 1119: Implement this logic on `FundingInfo` struct directly instead

tests/grpc.rs
```diff
@@ -247,6 +249,7 @@ async fn retry_until_bitcoin_funding_info(
         let request = tonic::Request::new(NeedsFundingRequest {
             id: 11,
             blockchain: farcaster::Blockchain::Bitcoin.into(),
+            network_selector: NetworkSelector::LocalNetworks.into(),
```
h4sh3d, line 252: FYI the other tests use `AllNetworks` and here you use `LocalNetworks`

h4sh3d (commented):
Some general comments and questions, otherwise all good!

src/grpcd/runtime.rs
TheCharlatan: No, we still need it. I think in case it's also useful in the cli. The call does take very long if you apply it to all network options.

tests/grpc.rs
TheCharlatan: Yes, that's intentional to test it.

src/grpcd/runtime.rs
TheCharlatan: I don't quite get what you want here.

h4sh3d: I think what I had in mind is wrong, this can stay as is.

src/grpcd/runtime.rs
```diff
@@ -1072,18 +1089,7 @@ impl Farcaster for FarcasterService {
...
-                                address: b_info.address.to_string(),
-                                amount: b_info.amount.as_sat(),
-                            },
-                            FundingInfo::Monero(m_info) => farcaster::FundingInfo {
-                                swap_id: m_info.swap_id.to_string(),
-                                address: m_info.address.to_string(),
-                                amount: m_info.amount.as_pico(),
-                            },
-                        })
+                        .map(|info| farcaster::FundingInfo::from(info))
```
h4sh3d, line 1092: Should be able to do `.map(farcaster::FundingInfo::from)`

h4sh3d: 49822589bc1c6f507de4ceb4ae0c8406b9d22008

h4sh3d (approved):

PR #821, opened 2022-12-13
Deps: Bump monero-rpc to latest version (v0.3.2)

TheCharlatan:
Use the new monero-rpc builder pattern for instantiating monero rpc clients.

### DIFF
--- a/src/syncerd/monero_syncer.rs
+++ b/src/syncerd/monero_syncer.rs
@@ -72,10 +72,23 @@ pub struct Transaction {
 }
 
 impl MoneroRpc {
-    fn new(node_rpc_url: String) -> Self {
+    fn new(node_rpc_url: String, proxy_url: Option<String>) -> Self {
+        let mut client_builder = monero_rpc::RpcClientBuilder::new();
+        if let Some(proxy_url) = proxy_url {
+            client_builder = client_builder.proxy_address(proxy_url);
+        }
+        let daemon_json_rpc = client_builder
+            .clone()
+            .build(node_rpc_url.clone())
+            .expect("client builder failed, cannot recover from bad configuration")
+            .daemon();
+        let daemon_rpc = client_builder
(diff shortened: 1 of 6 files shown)


src/syncerd/monero_syncer.rs
```diff
@@ -63,10 +63,23 @@ pub struct Transaction {
...
 
 impl MoneroRpc {
-    fn new(node_rpc_url: String) -> Self {
+    fn new(node_rpc_url: String, proxy_url: Option<String>) -> Self {
+        let mut client_builder = monero_rpc::RpcClientBuilder::new();
+        if let Some(proxy_url) = proxy_url {
+            client_builder = client_builder.proxy_address(proxy_url);
+        }
+        let daemon_json_rpc = client_builder
+            .clone()
```
h4sh3d, line 72: Can't we clone the client instead of the builder?

src/syncerd/monero_syncer.rs
```diff
@@ -63,10 +63,23 @@ pub struct Transaction {
 }
 
 impl MoneroRpc {
-    fn new(node_rpc_url: String) -> Self {
+    fn new(node_rpc_url: String, proxy_url: Option<String>) -> Self {
+        let mut client_builder = monero_rpc::RpcClientBuilder::new();
+        if let Some(proxy_url) = proxy_url {
```
h4sh3d, line 68: This is the only line in this patch that takes care of a potential proxy, just want to make sure that the rest is handled in #824 and we don't miss setting the proxy elsewhere.

src/syncerd/monero_syncer.rs
h4sh3d: Refactored in cb0d7afc5233abb36d634c362c6fcb8aed3b7f1e

h4sh3d (approved):

PR #824, opened 2022-12-13
Monero socks proxy

TheCharlatan:
Based on #821

### DIFF
--- a/src/syncerd/monero_syncer.rs
+++ b/src/syncerd/monero_syncer.rs
@@ -71,7 +71,7 @@ pub struct Transaction {
     block_hash: Option<Vec<u8>>,
 }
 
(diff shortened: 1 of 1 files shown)


src/syncerd/monero_syncer.rs
```diff
@@ -71,24 +71,22 @@ pub struct Transaction {
     block_hash: Option<Vec<u8>>,
 }
 
+fn create_monero_rpc_client(rpc_url: String, proxy_url: Option<String>) -> monero_rpc::RpcClient {
```
h4sh3d, line 74: As we are in `monero_syncer` module we could reduce this fn name to `create_rpc_client`

TheCharlatan: f72421341c872bf90b071a560e24abbc9872a62d

h4sh3d (approved):

PR #829, opened 2022-12-13
TSM: Remove unneeded peerd check

TheCharlatan:


h4sh3d (approved):
ACK. Functionally the same, the error log is not true.

PR #831, opened 2022-12-13
Syncer State Machine: Handle Task Aborted

TheCharlatan:
If a call to the syncer from Farcasterd fails, task aborted may be emitted. Handle this case gracefully.


Lederstrumpf (approved):
LGTM! :thumbsup:

Issue #271, opened 2021-11-30
btc and xmr amounts on syncerstate, find better place


zkao:

TheCharlatan:
Done with #838

PR #838, opened 2022-12-14
remove swapd.runtime.syncer_state.{bitcoin,monero}_amount

### LINKED ISSUE #837: don't track swap amount inside `swapd.runtime.syncer_state.{bitcoin, monero}_amount`
where this is needed, it can be read from runtime.public_offer

Lederstrumpf:
There's no reason to keep this within the syncer state since it's only
read from within the swap state machine. Instead, read it from `swapd.runtime.public_offer.offer.{arbitrating,according}_amount_`. Closes #837 and closes the more loosely phrased #271.


TheCharlatan (approved):
Easy win :)

PR #840, opened 2022-12-14
Alice: broadcast cancel tx eventually even if Alice overfunded

### LINKED ISSUE #834: Alice never broadcasts cancel if she's overfunded
If Alice overfunds, she moves on to AliceCanceled without ever broadcasting cancel.
https://github.com/thecharlatan/farcaster-node/blob/7eed93db585dbfa5a4f3df04ca38042030daa074/src/swapd/swap_state.rs#L1704-L1721

We should handle this inside [https://github.com/TheCharlatan/farcaster-node/blob/7eed93db585dbfa5a4f3df04ca38042030daa074/src/swapd/swap_state.rs#L1833-L1837](try_alice_canceled_to_alice_refund_or_alice_punish) by:

1. matching on `ArbLock` confirmations
2. checking if broadcasting cancel is safe
3. if cancel is safe, but Cancel's confirmation count is `None` (i.e. not on chain nor mempool), broadcasting cancel

This will also prepare for handling Cancel being reorged out.

Lederstrumpf:
Closes #834.

### DIFF
--- a/src/swapd/swap_state.rs
+++ b/src/swapd/swap_state.rs
@@ -1693,8 +1693,8 @@ fn try_alice_arbitrating_lock_final_to_alice_accordant_lock(
                     msg,
                 )?;
             } else if amount.clone() > required_funding_amount.as_pico() {
-                // Alice overfunded to ensure that she does not publish the buy transaction if Bob gives her the BuySig,
-                // go straight to AliceCanceled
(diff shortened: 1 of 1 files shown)


TheCharlatan (approved):
Rebase and autosquash plz

TheCharlatan (approved):

PR #842, opened 2022-12-15
If punish only end when punish final, and Bob check punish too

### LINKED ISSUE #833: Bob never gives up even if punished for realz
Bob currently plays ostrich and doesn't acknowledge being punished - he keeps watching for his refund to finalize (hoping to win the race) and continues running if not.
As pointed out by @h4sh3d, we should handle this exactly 1:1 with other finalization thresholds - there should be no difference in handling whether we like the outcome or not.
As such, once punish is final, Bob should transition to `SwapEnd(Outcome::FailurePunish)`.

Lederstrumpf:
Closes #833.

Bob doesn't know the txid of the punish tx, but he can infer it by virtue of watching the address of the output script of the `Cancel` tx for outgoing transactions that don't have the txid of `Refund`.

the best blocks available are added to each checkpoint, and when watching addresses, we now scan from that checkpoint or, if not available (only the case for the funding tx, the current block).


src/swapd/swap_state.rs
```diff
@@ -1374,7 +1374,7 @@ fn try_bob_cancel_final_to_swap_end(
             // send swap outcome to farcasterd
             Ok(Some(SwapStateMachine::SwapEnd(Outcome::FailureRefund)))
         }
-        _ => Ok(None),
+        _ => try_punish_to_swap_end(event, runtime),
```
TheCharlatan, line 1377: I'm not seeing where Bob is watching the Punish transaction. Does that still need to be added?

TheCharlatan (changes requested):

src/swapd/swap_state.rs
Lederstrumpf: it's working now on c1bcec836ddda10ce3ccd5038bff1a91f8dfe962 - I've added a description to this PR.

I've obviously merged in your work from #855 that enables filtering address transactions for IO direction, but also merged #863 to avoid unnecessary rewrites since it touches related code. My branch here is currently a mess, so will merge in both of yours first and then rebase & clean up this one so it can be reviewed.

Lederstrumpf: didn't rebase since it introduced some conflicts & created a broken history from overlapping changes with #863, but I've squashed commits where relevant to produce a reviewable history - assuming you skip all the merges.

src/swapd/swap_state.rs
```diff
@@ -2135,10 +2192,34 @@ fn attempt_transition_to_alice_reveal(
     }
 }
 
+fn bob_watch_cancel_output_task(runtime: &mut Runtime) -> Option<Task> {
```
TheCharlatan, line 2195: Implement this on the runtime directly, like other places where we mutate the Runtime.

tests/swap.rs
```diff
@@ -1998,19 +1999,43 @@ async fn run_punish_swap_kill_bob_before_monero_funding(
...
+    // wait a bit for Bob to get up to speed
+    tokio::time::sleep(time::Duration::from_secs(5)).await;
+
+    bitcoin_rpc
+        .generate_to_address(10, &reusable_btc_address())
+        .unwrap();
+    info!("generated 10 bitcoin blocks");
+
+    // wait a bit for Bob to get up to speed
+    tokio::time::sleep(time::Duration::from_secs(5)).await;
```
TheCharlatan, line 2029: This sleep is not needed, you are just waiting for the next transition and not issuing any commands from the tests.

tests/swap.rs
```diff
@@ -1998,19 +1999,43 @@ async fn run_punish_swap_kill_bob_before_monero_funding(
...
+    tokio::time::sleep(time::Duration::from_secs(2)).await;
+
+    restore_checkpoint(swap_id, data_dir_bob.clone());
+
+    // wait a bit for Bob to get up to speed
+    tokio::time::sleep(time::Duration::from_secs(5)).await;
+
+    bitcoin_rpc
+        .generate_to_address(10, &reusable_btc_address())
+        .unwrap();
```
TheCharlatan, line 2025: It's a bit weird that this is required. Shouldn't Bob automatically go through the transitions, since they are all already confirmed and final?

src/swapd/runtime.rs
```diff
@@ -867,6 +875,8 @@ impl Runtime {
                     connected_counterparty_node_id: get_node_id(&self.peer_service),
                     deal: self.deal.clone(),
                     monero_address_creation_height: self.monero_address_creation_height.clone(),
+                    bitcoin_checkpoint_height: self.syncer_state.bitcoin_height,
+                    monero_checkpoint_height: self.syncer_state.monero_height,
```
TheCharlatan, line 879: Why are these set with the checkpoints and not when the respective funding addresses are created?

src/swapd/runtime.rs
```diff
@@ -105,7 +105,9 @@ pub fn run(config: ServiceConfig, opts: Opts) -> Result<(), Error> {
         swap_id,
         tasks,
         monero_height: 0,
+        monero_checkpoint_height: None,
```
TheCharlatan, line 108: Is this for all intents and purposes the same as the `monero_address_creation_height`? If so, I'd refactor those two concepts together.

TheCharlatan (changes requested):

src/swapd/swap_state.rs
```diff
@@ -1293,35 +1293,92 @@ fn try_bob_cancel_final_to_swap_end(
...
+                    .tasks
+                    .txids
+                    .get(&TxLabel::Refund)
+                    .expect("should be watching Refund on CancelFinal")
+                && Some(txid)
+                    != runtime
+                        .syncer_state
+                        .tasks
+                        .txids
+                        .get(&TxLabel::Punish)
```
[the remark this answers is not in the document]
Lederstrumpf: if the `Punish` txid is already in `syncer_state.tasks`, we already watching this txid (because we've already hit this branch) and don't need to add another task.

src/swapd/swap_state.rs
Lederstrumpf: done in c9da29f628564a787638d76acb6263a7165fb223

src/swapd/swap_state.rs
```diff
@@ -1293,35 +1293,92 @@ fn try_bob_cancel_final_to_swap_end(
...
+            && !incoming =>
+        {
+            let txid = bitcoin::Txid::from_slice(&hash).expect("should serialize to valid txid");
+            if txid
+                != *runtime
+                    .syncer_state
+                    .tasks
+                    .txids
+                    .get(&TxLabel::Refund)
+                    .expect("should be watching Refund on CancelFinal")
```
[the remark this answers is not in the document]
Lederstrumpf: if we're in CancelFinal as Bob, we must have created a watch task for `Refund`, but only may have created one for `Punish`, which is why I distinguished them here.
But you're right that this isn't safe if due to some unexpected behavior the `Refund` task gets removed. I've changed this to a `debug_assert` invariant in 9d4eb95c5792846be1321cbd3a21fe6fd98f5b16 - then we'll still catch whether this assumption breaks during testing.

tests/swap.rs
Lederstrumpf: true - debugging remnant. removed in 34b53b9b8ad9420b761e3840d40d15417d419770.

tests/swap.rs
Lederstrumpf: 34b53b9b8ad9420b761e3840d40d15417d419770

src/swapd/runtime.rs
Lederstrumpf: when restoring a checkpoint, we don't need to scan back further than the last checkpoint for transactions that still matter for the swap state we're restoring to.

that being said, I think what we should be checkpointing is not the current block height, but subtract from it the finalisation threshold: In case of a reorg, we should expect up to our finalisation threshold of blocks to be removed, so a tx we're watching for (or txs related to an address we're watching) could be moved ahead by that many blocks, and we'd miss it when rewatching.

src/swapd/runtime.rs
TheCharlatan: Looks to me like we are not actually using `monero_checkpoint_height` in practice, but instead only the `monero_address_creation_height`. 

In terms of performance, the gains are marginal at best. Re-scanning a few blocks more in Monero's case hardly makes a difference and Bitcoin does not really care about height either way. We don't expect more than one transaction anyway from each of these addresses, so that should not be a concern here.

The checkpointed height for Bitcoin and Monero is also overwritten on each checkpoint made. If we re-watch any address, it won't return a previous transaction if it happened before the last checkpoint. While it's not a problem right now, since we don't need these transactions anymore, this feels a bit brittle.

Here's what I'd propose: We add a `HashMap<TxLabel, BlockHeight> for the address creation heights and insert whenever we create an address. Then we checkpoint and restore this data structure. What do you think?

TheCharlatan (approved):

PR #848, opened 2022-12-15
Service: Implement Reporter trait

TheCharlatan:
The Reporter trait is used to implement all report_ functions instead of inside the CtlServer trait. The Reporter allows an implementing service to define a target service to which the reports should be sent to.

In swapd, the only service where we use these functions, we define the enquirer as the service to report to.


h4sh3d (approved):
ACK

PR #851, opened 2022-12-15
Database: Cleanup Open Deals as well on CleanDanglingDeals

TheCharlatan:
As an alternative: We could re-add these open offers to the farcasterd state and launch the listener.

### DIFF
--- a/src/databased/runtime.rs
+++ b/src/databased/runtime.rs
@@ -178,9 +178,11 @@ impl Runtime {
                     .drain(..)
                     .map(|info| info.deal)
                     .collect();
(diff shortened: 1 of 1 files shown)


h4sh3d (approved):
cACK - I personally prefer this behavior to restarting listeners, if this is used in an exchange infrastructure the management/distribution of deals on nodes should take care of this, i.e. when you stop/restart a node either you know it's empty or the deals have to be recreated elsewhere.

What we could add is a `restore-deal`, this would allow externally moving a deal from a node to another.

PR #855, opened 2022-12-16
Syncer: Support Incoming and Outgoing AddressTransaction events

TheCharlatan:
Adds a filter to the WatchAddress Task in order for the user to select incoming or outgoing transactions. The returned AddressTransaction's amount is either the sum total incoming amount if the incoming bool is true, or the sum total outgoing amount if the incoming bool is false.

### DIFF
--- a/tests/monero.rs
+++ b/tests/monero.rs
@@ -8,11 +8,11 @@ use farcaster_node::syncerd::types::{
     Abort, AddressAddendum, Boolean, BroadcastTransaction, Task, WatchAddress, WatchHeight,
     WatchTransaction,
 };
-use farcaster_node::syncerd::GetAddressBalance;
 use farcaster_node::syncerd::{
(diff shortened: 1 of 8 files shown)


tests/monero.rs
```diff
@@ -334,6 +334,7 @@ async fn monero_syncer_address_test() {
                 lifetime: blocks + 1,
                 addendum: addendum_1.clone(),
                 include_tx: Boolean::True,
+                filter: TxFilter::Incoming,
```
Lederstrumpf, line 337: no tests of `TxFilter::Outgoing`, but this is fine since we don't need to watch outgoing txs with farcaster (we could check whether counterparty sweeps the xmr address, but not needed for the protocol).

Lederstrumpf (approved):
tACK @ c36a019a618eedc0cc1d15a730fe26f145bfa73d - nice work!

PR #863, opened 2022-12-17
Swap State Machine: Remove Alice Punish state

TheCharlatan:
Handle the punish as part of the AliceCanceled state

This fixes a bug, where Alice goes into AlicePunish state even though the Refund transaction was broadcasted and available.


Lederstrumpf (approved):
ACK @ d197a3d4567a26aafcc0a7765c09a8d0281558d1 & via c1bcec836ddda10ce3ccd5038bff1a91f8dfe962.

I didn't try to reproduce the restore bug since - from retracing the prior logic - in hindsight the bug now seems obvious from the segregation of the punish state (which didn't cater for a refund finalisation it may have missed prior to restoring), and handling both refund & punish finalisation possibilities from the same state is a much improved design.

PR #864, opened 2022-12-17
Errors: Wrap error types in Error enum instead of converting to Farcaster::Error

TheCharlatan:
Create a bunch of new error types and allign existing errors with common formatting.


h4sh3d (approved):

TheCharlatan (commented):
ACK

PR #865, opened 2022-12-17
Swapd: Handle re-orgs and possible mempool drops of broadcasted txs.

TheCharlatan:
Keep a list of broadcasted transactions and re-broadcast them when receiving a `None` confirmation event.

Also segregate the handling of syncer confirmations to the runtime and don't repeat the logic in the swap state machine.

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -116,6 +116,7 @@ pub fn run(config: ServiceConfig, opts: Opts) -> Result<(), Error> {
         awaiting_funding: false,
         xmr_addr_addendum: None,
         confirmations: none!(),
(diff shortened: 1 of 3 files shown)


Lederstrumpf (approved):
ACK @ 7a60e2d9f5fc510c47b75bec405e26b15c420dc5.

Started testing on fc.dev - iirc the only stuck swaps we had left boiled down to txs being broadcast and then nothing happening - so this may actually fix em :). Will check whether rebroadcasting addresses all of them. If it doesn't, means they never end up on the mempool, in which case we'll prolly have to cave for an uglier scheduled rebroadcaster ... or bribin' bitmain :D

PR #869, opened 2022-12-19
Chore: Handle some clippy findings

TheCharlatan:


.clippy.toml
```diff
@@ -1 +1,3 @@
 enum-variant-size-threshold = 4000
+too-many-arguments-threshold = 10
```
Lederstrumpf, line 2: naughty! we should be better about that ^^

Lederstrumpf (approved):

PR #872, opened 2022-12-20
Grpc: Handle handler drops (usually on client disconnect)

TheCharlatan:
When the client disconnects, its handler function is dropped. In this case, we should not error, but log a warning and continue the loop. Since we remove from the pending requests, no additional resources need to be dropped


h4sh3d (approved):
tACK ebcd5d9d33d132bc90c8a5d740b315fc4c0c2f22

PR #873, opened 2022-12-20
Syncer: Lower the time spent sleeping on the syncer receiver

TheCharlatan:
This should consume minimally more resources, while offering 10x faster processing times


h4sh3d (approved):
Woaw, can we scale blockchains 10x too?

PR #878, opened 2022-12-23
Cli: health check ignore local by default

### LINKED ISSUE #825: Disable local network in default health check command
We have a selector to query only 1 specific network which is required when building clients/gui that operate per network. However the cli health check command should disable by default the local and only do testnet and mainnet, as local is normally used only during tests.

h4sh3d:
Closes #825

By default no `local` configuration is provided, and local is only use in automated testing envs.

```
swap-cli health-check              <  check only mainnet and testnet
swap-cli health-check -s all       <  check all
swap-cli health-check -s {net}     < check only net
```

### DIFF
--- a/src/bus/types.rs
+++ b/src/bus/types.rs
@@ -273,8 +273,22 @@ impl FromStr for HealthCheckSelector {
     derive(Serialize, Deserialize),
     serde(crate = "serde_crate")
 )]
-#[display(HealthReport::to_yaml_string)]
-pub struct HealthReport {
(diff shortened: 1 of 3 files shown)


TheCharlatan (approved):
Nice change! If you feel like it, you can add a similar `check_health` helper function in the grpc runtime.

h4sh3d:
> Nice change! If you feel like it, you can add a similar `check_health` helper function in the grpc runtime.

I've open #880 to track this, but less important than this change IMO (it improves the codebase not the UX)

PR #881, opened 2022-12-23
Swapd: Cache all transaction related syncer events

TheCharlatan:
This allows us to replay the latest tx events for most transaction types.
We do this to immediately transition through states for which we already have all required information.


h4sh3d (approved):
tACK

PR #887, opened 2022-12-27
Swapd: Always abort all tasks on swap end

TheCharlatan:
Moves the responsibility of cleaning up everything on swap end from the swap state machine to the swap runtime.


TheCharlatan:
This also fixes a current build error on the main branch.

h4sh3d (approved):

PR #896, opened 2022-12-28
Swap: Add swap logging to syncer client

TheCharlatan:
Implements the extended logging currently used in the swapd Runtime as a Trait used by both the Runtime and the syncer client.


h4sh3d (approved):
tACK 1e7a2c5bd309776b6e77efb59b66f0abd820ddf6

PR #905, opened 2022-12-28
Revert "Ci: fix wiki branch"

h4sh3d:
Reverts farcaster-project/farcaster-node#902

This branch is for the repo not for the wiki, this was correct. This patch was not needed and breaks the sync.


h4sh3d:
I'm wondering what I did with this sync job... I guess the de-synchronization came from too much jobs in parallel while editing the wiki 2 weeks ago.

PR #908, opened 2022-12-28
Swapd: Simplify commitment creation

TheCharlatan:
Based on #906

### DIFF
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -859,6 +859,16 @@ impl Runtime {
             sweep_block.bright_blue_bold(),
         ));
     }
+
+    pub fn report_progress_message_no_fail(
+        &mut self,
+        endpoints: &mut Endpoints,
+        msg: impl ToString,
+    ) {
+        if let Err(err) = self.report_progress_message(endpoints, msg) {
+            self.log_error(format!("Error sending progress message: {}", err))
+        }
+    }
 }
 
 pub trait SwapLogging {
--- a/src/swapd/runtime.rs
+++ b/src/swapd/runtime.rs
@@ -860,7 +860,7 @@ impl Runtime {
         ));
     }
 
-    pub fn report_progress_message_no_fail(
+    pub fn report_progress_message_log_fail(
         &mut self,
         endpoints: &mut Endpoints,
         msg: impl ToString,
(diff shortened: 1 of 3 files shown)


src/swapd/runtime.rs
```diff
@@ -859,6 +859,16 @@ impl Runtime {
...
+
+    pub fn report_progress_message_no_fail(
+        &mut self,
+        endpoints: &mut Endpoints,
+        msg: impl ToString,
+    ) {
+        if let Err(err) = self.report_progress_message(endpoints, msg) {
+            self.log_error(format!("Error sending progress message: {}", err))
+        }
+    }
```
Lederstrumpf, line 871: nit: I'm guessing this is outside the reporter trait only because `.log_error` relies on the `SwapLogging` trait.
But the swapd runtime is the only one also implementing the `Reporter` trait.
So maybe move this into the reporter trait and add `SwapLogging` as a trait constraint? Or do you reckon we'll be using the `Reporter` trait outside the swap runtime too? Then it may make more sense to keep separate.

I also find the naming suffix here `_no_fail` not too descriptive (it doesn't return a result, yes, but name suggests it can't fail, yet it simply already "handles" the failure by logging it - maybe `_log_fail`?), but first wanna get your take on the above.

Lederstrumpf (approved):

src/swapd/runtime.rs
TheCharlatan:
```
nit: I'm guessing this is outside the reporter trait only because .log_error relies on the SwapLogging trait.
But the swapd runtime is the only one also implementing the Reporter trait.
So maybe move this into the reporter trait and add SwapLogging as a trait constraint? Or do you reckon we'll be using the Reporter trait outside the swap runtime too? Then it may make more sense to keep separate.
```

No, these should be separate. The goal is to refactor the `SwapLogging` trait and also apply it to the trade state machine. There is no `Reporter` in farcasterd. The `Reporter` trait is a leftover from lnp-node. In its current form, it makes little to no sense. If the SwapLogging trait is added as a constraint to the `Reporter` trait, it's clear that the `Reporter` will only ever be used in the swapd context. In that case, we might rather remove the `Reporter` and just send the progress messages in the same way we send all the others. Alternatively we implement the `Reporter` elsewhere, e.g. in walletd, where lack the context required for the `SwapLogging` trait.

```
I also find the naming suffix here _no_fail not too descriptive (it doesn't return a result, yes, but name suggests it can't fail, yet it simply already "handles" the failure by logging it - maybe _log_fail?), but first wanna get your take on the above.
```

Yes, `_log_fail` is much better.

TheCharlatan: 64464048b5e6f79ecdeed072a9712e17fa7508b4

h4sh3d (approved):

PR #910, opened 2022-12-28
Feat: cli enhancement and improved deal validation

h4sh3d:
Allows to express lower and upper bounds for each assets on each network, used when validating maker/taker deals.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -222,7 +222,7 @@ impl Config {
                     .map(|net| net.amounts)
             })
             .or_else(|| Self::btc_default_tradeable(deal.network))
-            .map(|tradeable| tradeable.map().validate_amount(deal.arbitrating_amount))
+            .map(|tradeable| tradeable.validate_amount(deal.arbitrating_amount))
             .transpose()?;
         // assume accordant is monero
         self.swap
@@ -233,7 +233,7 @@ impl Config {
                     .map(|net| net.amounts)
             })
             .or_else(|| Self::xmr_default_tradeable(deal.network))
-            .map(|tradeable| tradeable.map().validate_amount(deal.accordant_amount))
+            .map(|tradeable| tradeable.validate_amount(deal.accordant_amount))
             .transpose()?;
 
         Ok(())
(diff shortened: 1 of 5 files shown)


h4sh3d:
Local tests shows for now some weird behavior, do not merge yet.

h4sh3d:
This is now good for a review.  cc @TheCharlatan

src/config.rs
```diff
@@ -260,6 +398,72 @@ pub struct ArbConfig {
...
+    pub max_amount: Option<T>,
+}
+
+impl<T> TradeableAmounts<T>
+where
+    T: FromStr + Display,
+    T::Err: Display,
+{
+    /// Map tradeable amounts into new type
+    pub fn map<U>(self) -> TradeableAmounts<U>
```
TheCharlatan, line 427: What is this function for? I removed it and nothing broke.

src/config.rs
```diff
@@ -260,6 +398,72 @@ pub struct ArbConfig {
...
+            max_amount: self.max_amount.map(|max| max.into()),
+        }
+    }
+
+    /// Validate a given amount based on potential rules (min, max)
+    pub fn validate_amount(&self, amount: T) -> Result<(), Error>
+    where
+        T: PartialOrd<T> + Copy,
+    {
+        use config::ConfigError::Message;
```
TheCharlatan, line 444: Why is this imported here and nowhere else?

TheCharlatan (changes requested):

src/config.rs
h4sh3d: This function was needed before to convert internal types `btc::Amount` and `xmr::Amount` into real types from `bitcoin` and `monero` crates, after 16200a24e91b4859ee1b98435d0f75b021e0abfa this function was not needed anymore but I kept it as it is a generic implementation of `map` for the tradeable amounts type, so maybe someday we'll need it again.

src/config.rs
h4sh3d: It was imported at two places and not used everywhere. I import them at function scope level to shorten the error return lines, as this is used elsewhere it's now imported everywhere with c003b47c433a812edb3b23359e843a9555e39bb3

src/config.rs
```diff
@@ -189,6 +199,118 @@ impl Config {
...
+            Network::Mainnet => Some(Self::mainnet_btc_default_tradeable()),
+            _ => None,
+        }
+    }
+
+    // Helper function to return default mainnet btc tradeable amounts
+    fn mainnet_btc_default_tradeable() -> TradeableAmounts<bitcoin::Amount> {
+        use bitcoin::Amount;
+        TradeableAmounts {
+            min_amount: None,
```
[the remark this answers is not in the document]
h4sh3d: I think it is a good idea to add one now. I chosen something in the same range as Monero, `0.00001 BTC` which is currently `0.17 USD`.

Done in b1eafd4845858b39091658de95bdd21538d42d50

h4sh3d:
Thanks for the review @TheCharlatan, the result is now better (I took the opportunity to add more doc and fix my mistake in 19108957e3a642767237f7b8552faccac950a70a)

TheCharlatan (approved):

PR #925, opened 2023-01-10
Fix(tsm): swaps restoring

h4sh3d:
This PR addresses two issues:
- An `event` could affect multiple `tsms`
- When restoring swapd make sure to reuse peerd connection if in `registered_services`

d894fba40dbade85589895bd4e85f91e683f288b only changes `..request_to_tsm` and not `..request_to_ssm` as I don't see any cases where the same issue would apply. This patch would be way cleaner with stabilized `drain_filter` feature, but it's not so... I did an ugly fn :)

### DIFF
--- a/src/farcasterd/runtime.rs
+++ b/src/farcasterd/runtime.rs
@@ -965,68 +965,81 @@ impl Runtime {
         }
     }
 
-    fn match_request_to_trade_state_machine(
+    fn match_request_to_trade_state_machines(
         &mut self,
         req: &BusMsg,
         source: &ServiceId,
-    ) -> Result<Option<TradeStateMachine>, Error> {
+    ) -> Result<Option<Vec<TradeStateMachine>>, Error> {
+        // mutate `tsms` by removing all `tsm` matching the predicate and returning them
+        //
+        // we assume tsm are rather small entities and we can clone them even if
+        // we have a couple thousands of them
+        //
+        // only returns an array if at least one element match predicate
+        fn dummy_drain_filter<P: Fn(&TradeStateMachine) -> bool>(
+            tsms: &mut Vec<TradeStateMachine>,
+            predicate: P,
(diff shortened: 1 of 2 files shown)


src/farcasterd/runtime.rs
```diff
@@ -965,68 +965,81 @@ impl Runtime {
...
         source: &ServiceId,
-    ) -> Result<Option<TradeStateMachine>, Error> {
+    ) -> Result<Option<Vec<TradeStateMachine>>, Error> {
+        // mutate `tsms` by removing all `tsm` matching the predicate and returning them
+        //
+        // we assume tsm are rather small entities and we can clone them even if
+        // we have a couple thousands of them
+        //
+        // only returns an array if at least one element match predicate
+        fn dummy_drain_filter<P: Fn(&TradeStateMachine) -> bool>(
```
TheCharlatan, line 979: I think this can be done like this without cloning or draining.

```
.trade_state_machines
.iter()
.enumerate()
.filter_map(|(i, tsm)| {
    if tsm.swap_id() == Some(*swap_id) {
        Some(i)
    } else {
        None
    }
})
.map(|pos| self.trade_state_machines.remove(pos))
.collect::<Vec<TradeStateMachine>>()),
```

TheCharlatan (changes requested):

src/farcasterd/runtime.rs
h4sh3d: Re-reading the code and comment makes me realized that the comment is in fact incorrect: no clone happen in this function, we drain the vector and push back the non-matching tsms.

Comment fixed in 4f47091874ff808c754706575b32cf064645263c

TheCharlatan: ACK

TheCharlatan (approved):

PR #937, opened 2023-01-15
deal validation: only validate amount of outbound asset

### LINKED ISSUE #936: validate only the asset's deal amount which corresponds to swap role
A swap participant's limit for what they're willing to risk within a swap is agnostic to what the other side is risking in the trade - they only care about the inventory that they're risking themselves.

Thus:
a. xmr amount should only be validated if `SwapRole` is `Alice`
b. btc amount should only be validated if `SwapRole` is `Bob`

Lederstrumpf:
Fixes #936.
Running on farcaster.dev with limits `0.0001-0.01 BTC`, `0.1-100 XMR`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,6 +6,7 @@
 
 use config::ConfigError::Message;
 use farcaster_core::blockchain::Network;
+use farcaster_core::role::{SwapRole, TradeRole};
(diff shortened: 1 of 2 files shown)


TheCharlatan (approved):
ACK

PR #946, opened 2023-06-21
Syncers: Some address balance fixes

TheCharlatan:
* Calculate the the bitcoin address balance by adding both the confirmed and unconfirmed values.
* Print warn (and not debug) when balance retrieval fails.
* Correct log message when sending a bitcoin address balance task to a monero syncer.


h4sh3d (approved):
cACK

